PR #5, opened 2022-03-01
Add a reserved site ID (-2) for refresh key. Add method to keystore snapshot to return refresh key, fall back to master key if not found.

haitaottd:
Add a reserved site ID (-2) for refresh key. Add method to keystore snapshot to return refresh key, fall back to master key if not found.

### DIFF
--- a/src/main/java/com/uid2/shared/store/RotatingKeyStore.java
+++ b/src/main/java/com/uid2/shared/store/RotatingKeyStore.java
@@ -163,6 +163,12 @@ public class RotatingKeyStore implements IKeyStore, IMetadataVersionedStore {
         }
 
         @Override
+        public EncryptionKey getRefreshKey(Instant now) {
+            EncryptionKey refreshKey = this.getActiveSiteKey(Const.Data.RefreshKeySiteId, now);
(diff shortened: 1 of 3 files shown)


src/main/java/com/uid2/shared/store/RotatingKeyStore.java
```diff
@@ -162,6 +162,12 @@ public EncryptionKey getMasterKey(Instant now) {
             return this.getActiveSiteKey(Const.Data.MasterKeySiteId, now);
         }
 
+        @Override
+        public EncryptionKey getRefreshKey(Instant now) {
+            EncryptionKey refreshKey = this.getActiveSiteKey(Const.Data.RefreshKeySiteId, now);
+            return refreshKey != null ? refreshKey : this.getActiveSiteKey(Const.Data.MasterKeySiteId, now);
```
atarassov-ttd, line 168: Why is this fallback needed? This can be confusing/dangerous if something goes wrong.

haitaottd: This is for handling legacy key file that doesn't have refresh key. Removed as we'll coordinate deployment to ensure that doesn't happen.

atarassov-ttd:
LGTM

PR #6, opened 2022-03-17
Allow uid2 services to run e2e locally.

haitaottd:
* Use s3 endpoint to use localstack as S3 backend
* Tweak cloud/local path conversion for running on windows and localstack

### DIFF
--- a/src/main/java/com/uid2/shared/optout/OptOutCloudSync.java
+++ b/src/main/java/com/uid2/shared/optout/OptOutCloudSync.java
@@ -121,7 +121,7 @@ public class OptOutCloudSync implements ICloudSync {
 
     @Override
     public String toLocalPath(String path) {
-        if (path.startsWith("https://")) {
+        if (path.startsWith("http")) {
             try {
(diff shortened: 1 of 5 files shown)


src/main/java/com/uid2/shared/optout/OptOutCloudSync.java
```diff
@@ -337,11 +337,8 @@ private void localListFiles(ICloudStorage fsLocal, String dirToScan, String file
             } else if (OptOutUtils.getFileTimestamp(f) == null) {
                 LOGGER.warn("Unrecognized timestamp, unknown file " + f);
             }
-            if (Utils.IsWindows && f.indexOf(':') > 0) {
```
atarassov-ttd, line 340: What is the reason for removing windows specific stuff?

haitaottd: Actually I have to remove this to make it run on windows.

atarassov-ttd:
LGTM

PR #10, opened 2022-04-06
Allow a watcher to be set on UidCoreClient to subscribe response status of requests to core service

haitaottd:


src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -64,11 +65,16 @@
     private final boolean enforceHttps;
     private static boolean useSecureParameters = true;
     private boolean allowContentFromLocalFileSystem = false;
+    private AtomicReference<Handler<Integer>> responseWatcher;
 
     public static UidCoreClient createNoAttest(String attestationEndpoint, String userToken, ApplicationVersion appVersion, boolean enforceHttps) {
         return new UidCoreClient(attestationEndpoint, userToken, appVersion, CloudUtils.defaultProxy, new NoAttestationProvider(), enforceHttps);
     }
 
+    public static UidCoreClient createFakeForTest() {
```
atarassov-ttd, line 74: Is there a way to avoid putting unit test logic into prod code? Can this be solved by a mock?

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -110,6 +117,7 @@ private InputStream getWithAttest(String url) throws IOException, UidCoreClientE
     private boolean attestIfRequired(HttpURLConnection conn) throws IOException, UidCoreClientException {
         boolean attested = false;
         int statusCode = conn.getResponseCode();
+        notifyResponseStatusWatcher(statusCode);
```
atarassov-ttd, line 120: What is the reason for notifying response status here?

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -286,4 +296,14 @@ public void setPreSignedUrlExpiry(long expiry) {
     public String mask(String cloudPath) {
         throw new UnsupportedOperationException("UidCoreClient::mask method is not supported");
     }
+
+    public void setResponseStatusWatcher(Handler<Integer> watcher) {
```
atarassov-ttd, line 300: Should the watcher be specified in constructor? Is there a need for a setter?

src/main/java/com/uid2/shared/attest/UidCoreClient.java
haitaottd: moved to test code.

haitaottd: moved to test code.

src/main/java/com/uid2/shared/attest/UidCoreClient.java
haitaottd: In operator, UidCoreClient is created first, then shutdown handler is created during route setup of operator vertice.

haitaottd: In operator, UidCoreClient is created first, then shutdown handler is created during route setup of operator vertice.

src/main/java/com/uid2/shared/attest/UidCoreClient.java
haitaottd: All response codes are checked, not just attest requests. A malicious party could block core request on purpose.

src/main/java/com/uid2/shared/attest/UidCoreClient.java
atarassov-ttd: This does not have to be the case though.

haitaottd: I'd prefer to create shutdown handler inside `UIDOperatorVerticle` instead of creating it beforehand just for passing to UidCoreClient ctor.

haitaottd:
ping

tqinli (approved):
LGTM

PR #14, opened 2022-04-29
IDv3 support for optout

atarassov-ttd:
* add support for 33-byte advertising ID
* keep support for existing 32-byte advertising IDs


tqinli (approved):
LGTM

PR #17, opened 2022-05-13
Adding maven autobumb logic and workflows

cody-constine-ttd:


.github/workflows/check_version/action.yaml
```diff
@@ -0,0 +1,39 @@
...
+    description: 'The Version of the package'
+    value: ${{ steps.version.outputs.jar_version }}
+runs:
+  using: "composite" 
+
+  steps:
+    - name: Get Jar Version
+      id: version
+      shell: bash
+      run: echo "::set-output name=jar_version::$(mvn help:evaluate -Dexpression=project.version | grep -e '^[1-9][^\[]')"
```
atarassov-ttd, line 22: You could get rid of grep by adding `-q -DforceStdout` arguments to `mvn`.

PR #18, opened 2022-06-28
Added the encryptionhelpers and encrypted payload to shared

cody-constine-ttd:


IanAlexanderNara (approved):
Not able to access the classes through com.uid2.shared yet, guessing because not merged, but copied and pasted and everything works as expected. Your tests pass and my tests are all still passing after changing my code to the new class names/ method calls

haitaottd:
LGTM

PR #22, opened 2022-10-18
refactor UidCoreClient and add informative logging

scong-ttd:
Changes
- attestation process in UidCoreClient is separated in multiple helper functions without a good reason
- redact AttestationToken.ToString in case some exceptions print out values and call ToString implicitly
- attestation should always notify response status watcher, including the first one

### DIFF
--- a/src/main/java/com/uid2/shared/attest/UidCoreClient.java
+++ b/src/main/java/com/uid2/shared/attest/UidCoreClient.java
@@ -94,44 +94,43 @@ public class UidCoreClient implements IUidCoreClient, ICloudStorage {
         return this.contentStorage;
     }
 
-    public void attest() throws IOException, UidCoreClientException {
-        attestInternal();
-    }
-
+    /**
+     * HTTP Get with automatic attestation on first 401 Unauthorized response
+     * @param url API endpoint
+     * @return InputStream from HTTP response
+     * @throws IOException on HTTP failures
+     * @throws UidCoreClientException on attestation failures
+     */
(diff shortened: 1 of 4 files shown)


src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -94,44 +94,43 @@ public ICloudStorage getContentStorage() {
...
-
-        URLConnection conn = sendGet(url);
-
-        if (conn instanceof HttpURLConnection && attestIfRequired((HttpURLConnection) conn))
-            conn = sendGet(url);
-
-        return conn.getInputStream();
-    }
+        if (!attested()) {
+            try {
```
atarassov-ttd, line 106: Should this try/catch block be extracted into its own method?

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -94,44 +94,43 @@ public ICloudStorage getContentStorage() {
...
+     * @throws IOException on HTTP failures
+     * @throws UidCoreClientException on attestation failures
+     */
     private InputStream getWithAttest(String url) throws IOException, UidCoreClientException {
-        if (!attested())
-            attestInternal();
-
-        URLConnection conn = sendGet(url);
-
-        if (conn instanceof HttpURLConnection && attestIfRequired((HttpURLConnection) conn))
```
atarassov-ttd, line 107: I think this was preventing attestation in standalone dev mode. Does the standalone mode work now? What prevents attestation path in that mode?

PR #24, opened 2022-11-09
Vse UI d2 503 fix request capture metrics for v2 endpoints

vishalegbert-ttd:


cody-constine-ttd (approved):

src/main/java/com/uid2/shared/vertx/RequestCapturingHandler.java
```diff
@@ -75,8 +78,11 @@ private void capture(RoutingContext context, long timestamp, String remoteClient
 
         String path = null;
         try {
-            path = context.currentRoute().getPath();
-        } catch (NullPointerException ex) {
+            // If the current route is a known path, extract the full path from the request URI
+            if (context.currentRoute().getPath() != null) {
```
atarassov-ttd, line 82: What is the reason for this check?

vishalegbert-ttd: If the router doesn't recognise the path, it's null. There's some logic below that sets the metric's path tag to "unknown" if the path is null. So this helps preserve that existing logic

PR #28, opened 2022-12-15
Added site ID to operator key and updated tests

gmsdelmundo-ttd:


src/main/java/com/uid2/shared/auth/OperatorKey.java
```diff
@@ -1,7 +1,11 @@
 package com.uid2.shared.auth;
 
 import io.vertx.core.json.JsonObject;
+
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
+import java.util.stream.Collectors;
```
vishalegbert-ttd, line 8: Are these imports necessary?

sunnywu (approved):

PR #31, opened 2022-12-22
Uid2-598 differentiate public private operator

sunnywu:
Add a field in Operator Key to differentiate between public/private operator


src/main/java/com/uid2/shared/Utils.java
```diff
@@ -184,4 +183,22 @@ public static <T1, T2> int upperBound(List<T1> list, T2 value, BiPredicate<T2, T
         }
         return first;
     }
+
+    //can make this as an config in admin/core service config
+    public static final String SiteSpecificDataSubDirPath = "sites/";
```
aulme, line 188: I think in Java only classes should be named with PascalCase.
In the design spec I was going with /site rather than /sites so far, following more of a db table naming convention for "table" names to be singular.

src/main/java/com/uid2/shared/Utils.java
```diff
@@ -184,4 +183,22 @@ public static <T1, T2> int upperBound(List<T1> list, T2 value, BiPredicate<T2, T
...
+    public static String getSiteSpecificMetadataPathName(int siteId, String metadataPathName)
+    {
+        return SiteSpecificDataSubDirPath +siteId + metadataPathName;
+    }
+
+    public static String getMetadataPathName(boolean isPublicOperator, int siteId, String metadataPathName)
```
aulme, line 194: I'm thinking if this is  better done with an enum in case we have more different kinds of data in the future. This also directly reference operator in the name, which is more of an implementation detail, in reality this is site specific data. Eg in admin which will be reusing this nothing needs to make an assumption that an operator will be loading this data.

I suggest something like `enum DataScope { Global, Site }` (maybe can have a better name). Afterwards you could convert your boolean to an appropriate value in Core.

src/main/java/com/uid2/shared/Utils.java
sunnywu: ok will correct them after you have finalised the path to the site data directory paths and i can update this

src/main/java/com/uid2/shared/Utils.java
sunnywu: ok sure will change that

src/main/java/com/uid2/shared/Utils.java
sunnywu: Will do this in the new UID2-575 ticket branch later after UID2-598 is done

src/main/java/com/uid2/shared/Utils.java
sunnywu: Will do this in the new UID2-575 ticket branch later after UID2-598 is done

src/test/java/com/uid2/shared/auth/OperatorKeyTest.java
```diff
@@ -138,4 +141,72 @@ public void verifyRolesPropSetOperatorRoleAndOptoutRole() {
...
+        JsonObject jo = (JsonObject) Json.decodeValue(testJson);
+        OperatorKey c = OperatorKey.valueOf(jo);
+        Assert.assertTrue(c.getOperatorType() == OperatorType.PRIVATE);
+
+        c.setOperatorType(OperatorType.PUBLIC);
+        Assert.assertTrue(c.getOperatorType() == OperatorType.PUBLIC);
+    }
+
+    @Test
+    public void verifyIsPublicOperatorFlagOn() {
```
jingyi-gao-ttd, line 164: Since we don't really have a boolean field IsPublicOperator, would it be better to rename the tests here?

sunnywu: Thanks i have renamed these method names.

PR #33, opened 2023-01-03
Added roles to OperatorKey

jingyi-gao-ttd:
Changes:

- add roles to operator key
- add unit tests to make sure roles are optional


src/main/java/com/uid2/shared/auth/OperatorKey.java
```diff
@@ -77,12 +85,13 @@ public boolean equals(Object o) {
...
                 && this.contact.equals(b.contact)
                 && this.protocol.equals(b.protocol)
+                && this.roles.equals(this.roles)
                 && this.created == b.created;
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(key, name, contact, protocol, created);
+        return Objects.hash(key, name, contact, protocol, created, roles);
```
sunnywu, line 94: i tend to think while roles/siteid are OperatorKey's member variables, they can in theory change values  (public <=> private for example) unlike name/contact/key so i don't think roles should be used to generate hash code? any thoughts?

src/main/java/com/uid2/shared/auth/Roles.java
```diff
@@ -10,7 +10,7 @@
 
 public class Roles {
     public static <T extends Enum<T>> Set<T> getRoles(Class<T> type, JsonObject clientKeyJson) {
-        return getRoles(type, clientKeyJson.getJsonArray("roles"));
+        return getRoles(type, clientKeyJson.getJsonArray("roles", new JsonArray()));
```
sunnywu, line 13: What's the purpose of adding this param?

src/test/java/com/uid2/shared/auth/OperatorKeyTest.java
```diff
@@ -71,4 +71,34 @@ public void verifySiteIdPropIsOptionalForBackwardsCompatibility() {
...
+                "        \"contact\": \"admin@uid2.com\",\n" +
+                "        \"created\": 1617149276,\n" +
+                "        \"disabled\": false,\n" +
+                "        \"roles\": [ \"optout\" ]\n" +
+                "    }";
+
+        JsonObject jo = (JsonObject) Json.decodeValue(testJson);
+        OperatorKey c = OperatorKey.valueOf(jo);
+        Assert.assertTrue(c.getRoles().contains(Role.OPTOUT));
+    }
```
sunnywu, line 103: maybe add another couple of simple tests that defines an operator with just OPERATOR role, and another with both OPTOUT and OPERATOR roles as they are the most likely combos in prod ?

src/main/java/com/uid2/shared/auth/OperatorKey.java
```diff
@@ -26,14 +27,15 @@ public OperatorKey(String key, String name, String contact, String protocol, lon
...
-    public OperatorKey(String key, String name, String contact, String protocol, long created, boolean disabled, Integer siteId) {
+    public OperatorKey(String key, String name, String contact, String protocol, long created, boolean disabled, Integer siteId, Set<Role> roles) {
         this.key = key;
         this.name = name;
         this.contact = contact;
         this.protocol = protocol;
         this.created = created;
         this.disabled = disabled;
         this.siteId = siteId;
+        this.roles = Collections.unmodifiableSet(roles);
```
sunnywu, line 38: are you going to require user to explicitly specify Role.OPERATOR in `roles` param (in admin service)? or is it better to hardcode the Role.OPERATOR here as Operator must have Role.OPERATOR anyway?

src/main/java/com/uid2/shared/auth/OperatorKey.java
```diff
@@ -26,14 +27,15 @@ public OperatorKey(String key, String name, String contact, String protocol, lon
         this.siteId = null;
```
sunnywu, line 27: might want to hardcode Role.OPERATOR here for backward compatibility - the shared lib might be pushed to prod first while using old Core/Admin service versions so need to consider admin service might not be calling the new constructor below but this one.

src/main/java/com/uid2/shared/auth/OperatorKey.java
sunnywu: for backward compatibility reason again, might want to add another constructor instead of just adding the `roles` param to this constructor. again might want to hardcode Role.OPERATOR in this roles-less constructor call

src/main/java/com/uid2/shared/auth/OperatorKey.java
jingyi-gao-ttd: Ah, yes I agree, I will take it out

src/main/java/com/uid2/shared/auth/Roles.java
jingyi-gao-ttd: this is for backward compatitble, the new JsonArray() is the default value,  if there is no "roles" field in the JSON, this returns an empty json array. Otherwise clientKeyJson.getJsonArray("roles") will throw an exception

src/main/java/com/uid2/shared/auth/OperatorKey.java
jingyi-gao-ttd: okay, got it

src/main/java/com/uid2/shared/auth/OperatorKey.java
jingyi-gao-ttd: Currently, in admin service I was using the first constructor which not required to set siteId and roles, and chain it with withRoles if we want to add a role to the optout key. So I am not sure if we need another constructor? But I totally agree we should make Role.OPERATOR to be the default role

src/test/java/com/uid2/shared/auth/OperatorKeyTest.java
jingyi-gao-ttd: Tests added

src/main/java/com/uid2/shared/auth/OperatorKey.java
sunnywu: i meant when you make the respective change in admin service (which will allow setting additional OPTOUT role), it should call such "another constructor" instead of extending the existing one like above, for backward compatibility.

jingyi-gao-ttd: Oh, I see what you mean, I will add the new constructor

src/test/java/com/uid2/shared/auth/OperatorKeyTest.java
sunnywu: might wanna test the `public boolean hasRole(Role role)` calls in these 2 tests as well instead of just testing `getRoles`

src/test/java/com/uid2/shared/auth/OperatorKeyTest.java
```diff
@@ -71,4 +71,68 @@ public void verifySiteIdPropIsOptionalForBackwardsCompatibility() {
...
+                "        \"name\": \"admin@uid2.com\",\n" +
+                "        \"contact\": \"admin@uid2.com\",\n" +
+                "        \"created\": 1617149276,\n" +
+                "        \"disabled\": false,\n" +
+                "        \"roles\": [ \"operator\" ]\n" +
+                "    }";
+
+        JsonObject jo = (JsonObject) Json.decodeValue(testJson);
+        OperatorKey c = OperatorKey.valueOf(jo);
+        Assert.assertTrue(c.getRoles().contains(Role.OPERATOR));
```
sunnywu, line 120: wanna test !hasRoles(Role.OPTOUT) ?

src/test/java/com/uid2/shared/auth/OperatorKeyTest.java
```diff
@@ -71,4 +71,68 @@ public void verifySiteIdPropIsOptionalForBackwardsCompatibility() {
...
+                "        \"name\": \"admin@uid2.com\",\n" +
+                "        \"contact\": \"admin@uid2.com\",\n" +
+                "        \"created\": 1617149276,\n" +
+                "        \"disabled\": false,\n" +
+                "        \"roles\": [ \"operator\", \"optout\" ]\n" +
+                "    }";
+
+        JsonObject jo = (JsonObject) Json.decodeValue(testJson);
+        OperatorKey c = OperatorKey.valueOf(jo);
+        Assert.assertTrue(c.getRoles().contains(Role.OPERATOR));
```
sunnywu, line 136: wanna test hasRoles(Role.OPTOUT) ?

src/test/java/com/uid2/shared/auth/OperatorKeyTest.java
```diff
@@ -71,4 +71,68 @@ public void verifySiteIdPropIsOptionalForBackwardsCompatibility() {
...
+                "        \"key\": \"test-admin-key\",\n" +
+                "        \"name\": \"admin@uid2.com\",\n" +
+                "        \"contact\": \"admin@uid2.com\",\n" +
+                "        \"created\": 1617149276,\n" +
+                "        \"disabled\": false\n" +
+                "    }";
+
+        JsonObject jo = (JsonObject) Json.decodeValue(testJson);
+        OperatorKey c = OperatorKey.valueOf(jo);
+        Assert.assertTrue(c.getRoles().contains(Role.OPERATOR));
```
sunnywu, line 87: wanna test !hasRoles(Role.OPTOUT) ?

src/main/java/com/uid2/shared/auth/OperatorKey.java
```diff
@@ -77,6 +96,7 @@ public boolean equals(Object o) {
                 && this.name.equals(b.name)
                 && this.contact.equals(b.contact)
                 && this.protocol.equals(b.protocol)
+                && this.roles.equals(this.roles)
```
sunnywu, line 99: @gmsdelmundo-ttd you didn't check for siteId here previously - is there any reason not to?

src/main/java/com/uid2/shared/auth/OperatorKey.java
```diff
@@ -34,6 +35,18 @@ public OperatorKey(String key, String name, String contact, String protocol, lon
...
+    public OperatorKey(String key, String name, String contact, String protocol, long created, boolean disabled, Integer siteId, Set<Role> roles) {
+        this.key = key;
+        this.name = name;
+        this.contact = contact;
+        this.protocol = protocol;
+        this.created = created;
+        this.disabled = disabled;
+        this.siteId = siteId;
+        roles.add(Role.OPERATOR);
+        this.roles = Collections.unmodifiableSet(roles);
```
gmsdelmundo-ttd, line 49: Since we should be using Java 11, `Set.of()` might be a better choice than `Collections.unmodifiableSet()` since the former doesn't allow null values.

src/main/java/com/uid2/shared/auth/OperatorKey.java
gmsdelmundo-ttd: I saw that the `disabled` flag was also not included, so I initially didn't add it. On second thought now, it should be added as well. @sunnywu Should we add both `disabled` and `siteId` to the `equals()` comparison?

Btw @jingyi-gao-ttd If you change the `equals()` comparison, you also have to update `hashCode()` with the same fields as well.

src/main/java/com/uid2/shared/auth/Roles.java
gmsdelmundo-ttd, line 13: Would it be better to pass the "roles" field directly into this rather than the whole `JsonObject`?

src/main/java/com/uid2/shared/auth/OperatorKey.java
sunnywu: @gmsdelmundo-ttd i think it depends on what we use the equals()/hashCode() for. So say if we want to confirm every field is equal then sure we should add both `disabled` and `siteId` in the equals comparison. But what about using the methods in a Set where we want to check if an OperatorKey object is in it but not necessarily caring the attribute like disabled/roles?

sunnywu: by just comparing the key/name/contact which has to be unique combo? 

but i tend to agree we should add all fields to the equals comparison and the Set usage is an edge case ?

src/main/java/com/uid2/shared/auth/OperatorKey.java
```diff
@@ -43,8 +56,13 @@ public OperatorKey(String key, String name, String contact, String protocol, lon
     public long getCreated() { return created; }
     public boolean isDisabled() { return disabled; }
     public void setDisabled(boolean disabled) { this.disabled = disabled; }
+    public Set<Role> getRoles() {
+        return roles;
+    }
+
     public Integer getSiteId() { return siteId; }
     public void setSiteId(Integer siteId) { this.siteId = siteId; }
+    public OperatorKey withRoles(Set<Role> roles) { this.roles = Collections.unmodifiableSet(roles); return this; }
```
gmsdelmundo-ttd, line 65: Might be better to just add a setter like the other fields:
`public void setRoles(Set<Role> roles) { ... }`

src/main/java/com/uid2/shared/auth/OperatorKey.java
gmsdelmundo-ttd: As per our discussion, let's just include all fields in hashCode and equals here

src/main/java/com/uid2/shared/auth/OperatorKey.java
jingyi-gao-ttd: @gmsdelmundo-ttd I think Set.of() only takes a collection of elements but the roles here is a Set, could we use Set.of?

src/main/java/com/uid2/shared/auth/OperatorKey.java
jingyi-gao-ttd: I've added disabled, siteId and roles in hashCode and equals

src/main/java/com/uid2/shared/auth/Roles.java
jingyi-gao-ttd: There are two `getRoles`, the second one we are calling from here is just taken the roles field, this is the wrapper, should we remove this wrapper but calling `getRoles(type, clientKeyJson.getJsonArray("roles", new JsonArray()));` from OperatorKey and ClientKey?

sunnywu (approved):

PR #34, opened 2023-01-05
Add site ID to http_requests counter

mcollins-ttd:


src/main/java/com/uid2/shared/vertx/RequestCapturingHandler.java
```diff
@@ -181,14 +185,32 @@ private void capture(RoutingContext context, long timestamp, String remoteClient
...
-    private void incrementMetricCounter(String apiContact, String host, int status, HttpMethod method, String path) {
+    private static Integer getSiteId(RoutingContext context) {
+        final Integer siteId = context.get(Const.RoutingContextData.SiteId);
+        if (siteId != null) {
+            return siteId;
+        }
+
+        final IAuthorizable profile = AuthMiddleware.getAuthClient(context);
+        if (profile instanceof OperatorKey) {
+            return ((OperatorKey) profile).getSiteId();
```
vishalegbert-ttd, line 196: I think we should add the getSiteId to the IAuthorizable interface

mcollins-ttd: Done 👍

PR #36, opened 2023-01-09
Adding flexibility to data providers to allow reading from different scopes - global or site

aulme:


src/main/java/com/uid2/shared/auth/RotatingKeyAclProvider.java
```diff
@@ -1,119 +0,0 @@
...
-        return aclsSpec.size();
-    }
-
-    public void loadContent() throws Exception {
-        this.loadContent(this.getMetadata());
-    }
-
-    @Override
-    public AclSnapshot getSnapshot(Instant asOf) {
-        return this.latestSnapshot.get();
```
sunnywu, line 87: this old code is rather puzzling - taking an time input but always return the latest copy

src/main/java/com/uid2/shared/store/scope/StoreScope.java
```diff
@@ -0,0 +1,11 @@
+package com.uid2.shared.store.scope;
+
+import com.uid2.shared.store.CloudPath;
+
+public interface StoreScope {
+    Integer getId();
+    CloudPath getMetadataPath();
+    CloudPath resolve(CloudPath cloudPath);
+
```
gmsdelmundo-ttd, line 9: No newline here for consistency

src/main/java/com/uid2/shared/store/scope/SiteScope.java
```diff
@@ -0,0 +1,30 @@
...
+import com.uid2.shared.store.CloudPath;
+
+public class SiteScope implements StoreScope {
+    public SiteScope(CloudPath rootMetadataPath, Integer siteId){
+        this.rootMetadataPath = rootMetadataPath;
+        this.siteId = siteId;
+    }
+    private final Integer siteId;
+    private final CloudPath rootMetadataPath;
+
```
gmsdelmundo-ttd, line 13: Remove newline

src/main/java/com/uid2/shared/store/reader/RotatingKeyStore.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Override
+    public long loadContent(JsonObject metadata) throws Exception {
+        return reader.loadContent(metadata, "keys");
+    }
+
+    public void loadContent() throws Exception {
+        this.loadContent(this.getMetadata());
+    }
+
```
gmsdelmundo-ttd, line 81: Remove space here for consistency

src/main/java/com/uid2/shared/store/CloudPath.java
```diff
@@ -0,0 +1,57 @@
...
+        CloudPath cloudPath = (CloudPath) o;
+        return path.equals(cloudPath.path);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(path);
+    }
+
+    private static final boolean isWindows = System.getProperty("os.name").toLowerCase().contains("win");
```
gmsdelmundo-ttd, line 56: Move this to the top of the class

src/main/java/com/uid2/shared/store/CloudPath.java
```diff
@@ -0,0 +1,57 @@
...
+    public String toString() {
+        String path = this.path.toString();
+
+        // Java's Path always converts everything to path standards of the OS it's running on.
+        // The paths are actually S3 paths so should always be Linux style, so we need to convert them when on Windows
+        return isWindows ? path.replace("\\", "/") : path ;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
gmsdelmundo-ttd, line 44: Include `isWindows` in the check (and update hashCode as well)

src/main/java/com/uid2/shared/store/CloudPath.java
aulme: Is that a Java convention?

src/main/java/com/uid2/shared/store/FileName.java
```diff
@@ -0,0 +1,24 @@
+package com.uid2.shared.store;
+
+public class FileName {
+    private final String prefix;
+    private final String suffix;
```
gmsdelmundo-ttd, line 5: Does suffix here refer to the filename itself? Or is prefix the filename and suffix the file extension?

aulme: Yes, suffix is the extension including the dot.

src/main/java/com/uid2/shared/store/FileName.java
```diff
@@ -0,0 +1,24 @@
+package com.uid2.shared.store;
+
+public class FileName {
```
sunnywu, line 3: is this intentionally unused? doesn't seem to be used in admin

src/main/java/com/uid2/shared/store/FileName.java
gmsdelmundo-ttd: Could you add a comment there specifying that?

src/main/java/com/uid2/shared/store/reader/RotatingKeyAclProvider.java
```diff
@@ -0,0 +1,50 @@
...
+    @Override
+    public AclSnapshot getSnapshot(Instant asOf) {
+        return reader.getSnapshot();
+    }
+
+    @Override
+    public AclSnapshot getSnapshot() {
+        return this.getSnapshot(Instant.now());
+    }
+
```
gmsdelmundo-ttd, line 49: Remove newline here for consistency

src/main/java/com/uid2/shared/store/CloudPath.java
aulme: Why would it be in check. It's a static, not specific to any instance.

src/main/java/com/uid2/shared/store/reader/RotatingKeyAclProvider.java
```diff
@@ -0,0 +1,50 @@
...
+        return reader.loadContent(metadata, "keys_acl");
+    }
+
+    public void loadContent() throws Exception {
+        this.loadContent(this.getMetadata());
+    }
+
+    @Override
+    public AclSnapshot getSnapshot(Instant asOf) {
+        return reader.getSnapshot();
```
gmsdelmundo-ttd, line 42: Do we ever make use of the `asOf` parameter?

src/main/java/com/uid2/shared/store/CloudPath.java
gmsdelmundo-ttd: Oh yes, I missed the static, nvm

src/main/java/com/uid2/shared/store/reader/RotatingKeyStore.java
aulme: Ok

src/main/java/com/uid2/shared/store/scope/SiteScope.java
aulme: Ok

src/main/java/com/uid2/shared/store/scope/StoreScope.java
aulme: Ok

src/main/java/com/uid2/shared/store/FileName.java
aulme: It is used a lot, e.g. ScopedStoreWriter

src/main/java/com/uid2/shared/store/parser/ParsingResult.java
```diff
@@ -0,0 +1,19 @@
+package com.uid2.shared.store.parser;
+
+public class ParsingResult<T> {
+    private final T data;
+    private final Integer count;
```
gmsdelmundo-ttd, line 5: Do we need to keep track of the `count`?

src/main/java/com/uid2/shared/store/ScopedStoreReader.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+public class ScopedStoreReader<T> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ScopedStoreReader.class);
+    private final ICloudStorage metadataStreamProvider;
+    private final StoreScope scope;
+    private final Parser<T> parser;
+    private final String dataTypeName;
+    private final ICloudStorage contentStreamProvider;
+    private final AtomicReference<T> latestSnapshot = new AtomicReference<T>();
```
gmsdelmundo-ttd, line 25: You can remove the `T` in `new`
i.e. `private final AtomicReference<T> latestSnapshot = new AtomicReference<>();`

src/main/java/com/uid2/shared/store/FileName.java
aulme: Ah you mean that the file is in Shared and all usages are in Admin, good point. I think it will probably end up used in Shared before the end of this story. Happy to move it to Admin for now though.

src/main/java/com/uid2/shared/store/CloudPath.java
aulme: Ok, done

src/main/java/com/uid2/shared/store/FileName.java
sunnywu: ScopedStoreWriter.java isn't included in this PR. Happy for it to be in Admin/Shared but if it's used by Admin for now then maybe Admin is best (saved you updating shared lib unnecessarily)

src/main/java/com/uid2/shared/store/reader/RotatingKeyAclProvider.java
aulme: Don't think so, this behavior is migrated from the old implementation. I guess this is prep for a change where we'll do?

src/main/java/com/uid2/shared/store/ScopedStoreReader.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+    private long loadContent(String path) throws Exception {
+        final InputStream inputStream = this.contentStreamProvider.download(path);
+        final ParsingResult<T> parsed = parser.deserialize(inputStream);
+        this.latestSnapshot.set(parsed.getData());
+        LOGGER.info(String.format("Loaded %d %s", parsed.getCount(), dataTypeName));
+        return parsed.getCount();
+    }
+
+    public long loadContent(JsonObject metadata, String dataType) throws Exception {
```
gmsdelmundo-ttd, line 64: Would you need to do `null` checks for all the `loadContent()` that consumes the `JsonObject metadata` directly? Since `getMetadata()` may potentially return null

src/main/java/com/uid2/shared/store/reader/RotatingKeyAclProvider.java
aulme: Ok

src/main/java/com/uid2/shared/store/reader/RotatingKeyAclProvider.java
gmsdelmundo-ttd: @sunnywu I think this is similar to the question you had?

src/main/java/com/uid2/shared/store/FileName.java
aulme: Ok, will move to Admin

src/main/java/com/uid2/shared/store/reader/RotatingKeyAclProvider.java
sunnywu: yeh that's from old code before Aleks' changes. let's keep it that way for now...

src/main/java/com/uid2/shared/store/ScopedStoreReader.java
aulme: ok

src/main/java/com/uid2/shared/store/reader/RotatingKeyAclProvider.java
gmsdelmundo-ttd: Sure

src/main/java/com/uid2/shared/store/ScopedStoreReader.java
aulme: Good point, a good question what we should do if it is null (ie there is no metadata). Probably the appropriate thing is to throw a better exception than a null pointer exception since if we get null from getMetadata we shouldn't really loadContent and should create it instead.

src/main/java/com/uid2/shared/store/parser/ParsingResult.java
aulme: Yes since all the store methods return the count. Not sure why is that useful though - possibly can get rid of it if nobody needs it - hard to tell with changes in Shared though.

src/main/java/com/uid2/shared/store/ScopedStoreReader.java
gmsdelmundo-ttd: If we go with the approach of creating when it's null, I guess we don't need to throw an exception right?

src/test/java/com/uid2/shared/auth/RotatingKeyAclProviderTest.java
```diff
@@ -27,7 +30,7 @@ public class RotatingKeyAclProviderTest {
 
     @Before public void setup() {
         mocks = MockitoAnnotations.openMocks(this);
-        keyAclProvider = new RotatingKeyAclProvider(cloudStorage, "metadata");
+        keyAclProvider = new RotatingKeyAclProvider(cloudStorage, new GlobalScope(new CloudPath("metadata")));
```
sunnywu, line 33: is it worth adding a separate test for using new SiteScope and also that would test the other overloaded `com.uid2.shared.store.reader.RotatingKeyAclProvider#loadContent()` method (not just `loadContent(JsonObject metadata`) ) for complete coverage?

src/main/java/com/uid2/shared/store/ScopedStoreReader.java
aulme: It should happen but nothing guarantees that this is how it will go. Exceptions are exactly for that - when things go not in the way you expect them to go. Additionally only Admin would do the writing - if another service encounters this situation, if eg Core got pointed to an empty directory, they can't do more than error.

src/test/java/com/uid2/shared/auth/RotatingKeyAclProviderTest.java
aulme: Ok, will add a test with site scope. Will also add one for loadContent().
Honestly, maybe we should get rid of loadContent(metadata), so far I've never seen anything using anything but a.loadContent(a.getMetadata()) which is the exact implementation of the loadContent()

aulme: Ok, done.

src/main/java/com/uid2/shared/store/ScopedStoreReader.java
aulme: Ok resolved and added testing around the whole area.

PR #40, opened 2023-01-10
Gcp add metadata field check

scong-ttd:
Changes
1. disallow all metadata fields except "user-data" to improve security
2. small refactoring to avoid implicit null semantic and provide clearer failure reasons

### DIFF
--- a/src/main/java/com/uid2/shared/secure/gcp/VmConfigId.java
+++ b/src/main/java/com/uid2/shared/secure/gcp/VmConfigId.java
@@ -25,13 +25,23 @@ public class VmConfigId {
         return this.isValid;
     }
 
-    public String getValue() {
-        if (!isValid) throw new RuntimeException("value is invalid - check isValid() before using VmConfigId");
+    /**
+     * Get string representation of the vmConfigId
+     * @return vmConfigId string
+     * @throws AssertionError if the value is inValid. Check isValid() before calling.
+     */
+    public String getValue() throws AssertionError {
+        assert isValid : "check isValid() before getting the value";
         return idString;
     }
 
-    public String getFailedReason() {
-        if (isValid) throw new RuntimeException("no failed reason - check isValid() before using VmConfigId");
+    /**
+     * Get why we did not create a vmConfigId successfully
+     * @return reason it failed
+     * @throws AssertionError if we have a valid vmConfigId. Check isValid() before calling.
+     */
+    public String getFailedReason() throws AssertionError {
+        assert !isValid : "check isValid() before getting failed reason";
         return failedReason;
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/uid2/shared/secure/gcp/VmConfigId.java
```diff
@@ -0,0 +1,51 @@
...
+        this.failedReason = reason;
+        this.projectId = projectId;
+    }
+
+    public boolean isValid() {
+        return this.isValid;
+    }
+
+    public String getValue() {
+        if (!isValid) throw new RuntimeException("value is invalid - check isValid() before using VmConfigId");
```
thomasm-ttd, line 29: Not sure of the java conventions, but throwing an exception in a getter could be very confusing for a caller. I would not expect a getter to ever throw an exception. Same for getFailedReason

scong-ttd: Comparing to a normal data type, this class is meant to carry the result of an attempt to generate VmConfigId (like Optional<String>), so the promoted usage is
```
final VmConfigId id = generateVmConfigId();
if(vmConfigId.isValid()) print(id.getValue());
else print(id.getFailedReason());
```
Calling the getter without checking validity is bad code and I want it to shout as loudly as possible :)

thomasm-ttd: I have had designs like this bite me hard in the past. What happens for example if the instance is passed to a logger that just calls the getters? What would be wrong with returning either an empty string, or a known value such as "IsValid not called, invalid result" or some other string?

scong-ttd: If it is used incorrectly and we returned null or invalid value, we are allowing invalid values to be passed into business logic and can cause wrong behaviors (for example, passing the attestation when it should not). The overriden "toString" method did reveal all the fields for logging purposes.

scong-ttd: Or put it another way: I want to make sure user checks "isValid" before using the idString, if there a compile-time way to enforce it I'm more than happy to do so 🤔

thomasm-ttd: Then can we change it to a method rather than a getter?

scong-ttd: I can't think of a proper name because the purpose of the method is getting an Optional value, and as Optional.get can throw NoSuchElementException I'm consider throwing fine.

Also I've changed it to assertion instead of runtimeException to indicate this is a logic error in code.

scong-ttd: removed the assertions. I think the problem I'm trying to solve is the caller ignores or forgets to check the return state, throwing is not the best way to deal with it but a static analysis tool (which is out of the scope of the change).

PR #41, opened 2023-01-11
update attestation token to use AES/GCM/NoPadding

scong-ttd:
Update attestation token encryption algorithm to AES/GCM/NoPadding - first half (UID2-538)

Reason for implementing backward compatibility: optout does not generate new ATs but needs to validate ATs generated by UID Core.

After making sure optout and core have both been updated to new algorithm we will clean up the code.

### DIFF
--- a/src/main/java/com/uid2/shared/attest/AttestationToken.java
+++ b/src/main/java/com/uid2/shared/attest/AttestationToken.java
@@ -24,6 +24,11 @@ public class AttestationToken {
     private final long expiresAt;
     private final long nonce;
     private final boolean isValid;
+    private static final int GCM_AUTHTAG_LENGTH_BYTE = 16;
+
+    // The AES-GCM specification recommends that the IV should be 96 bits long
+    // https://developer.mozilla.org/en-US/docs/Web/API/AesGcmParams
+    private static final int GCM_IV_LENGTH = 12;
 
     public AttestationToken(String plaintext, Instant expiresAt) {
         this(plaintext, expiresAt.getEpochSecond(), generateNonce(), true);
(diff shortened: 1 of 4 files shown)


src/main/java/com/uid2/shared/attest/AttestationToken.java
```diff
@@ -50,26 +49,67 @@ public static AttestationToken fromPlaintext(String plaintext) {
...
+
+    @Deprecated
+    private static String decryptOld(byte[] cipherText, byte[] iv, String paraphrase, String salt) throws Exception {
+        Cipher cipher = Cipher.getInstance("AES/CBC/PKCS5Padding");
+        cipher.init(Cipher.DECRYPT_MODE, getKeyFromPassword(paraphrase, salt), new IvParameterSpec(iv));
+        byte[] plaintext = cipher.doFinal(cipherText);
+        return new String(plaintext);
+    }
+
+    // TODO: replace encode with encodeNew
```
thomasm-ttd, line 91: Is the TODO still necessary?

scong-ttd: Yes, there will be a follow-up PR
- enabling GCM as the encrypting method
- removing old methods

src/main/java/com/uid2/shared/attest/AttestationToken.java
```diff
@@ -104,6 +144,12 @@ private static IvParameterSpec generateIv() {
         return new IvParameterSpec(iv);
     }
 
+    private static GCMParameterSpec generateGcmParam() {
+        byte[] iv = new byte[16];
```
atarassov-ttd, line 148: I believe this should 12. Compare for example uid2-operator v3 token encryption.

scong-ttd: good call. updated 👍

atarassov-ttd (approved):

PR #44, opened 2023-01-13
Sort the roles in alphabetic order

jingyi-gao-ttd:
Changes:
- Sort roles in alphabetic order to avoid unnecessary changes to the role list
- Fix nullpointerexception when constructor take null as roles
- Check the existence of roles key in the JSON object first rather than always create an empty object as default value

### DIFF
--- a/src/main/java/com/uid2/shared/auth/OperatorKey.java
+++ b/src/main/java/com/uid2/shared/auth/OperatorKey.java
@@ -52,8 +52,7 @@ public class OperatorKey implements IRoleAuthorizable<Role> {
         this.created = created;
         this.disabled = disabled;
         this.siteId = siteId;
-        roles.add(Role.OPERATOR);
-        this.roles = Collections.unmodifiableSet(roles);
+        this.roles = this.reorderAndAddDefaultRole(roles);
         this.operatorType = defaultOperatorType;
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/uid2/shared/auth/OperatorKey.java
```diff
@@ -105,6 +102,13 @@ public static OperatorKey valueOf(JsonObject json) {
                 );
     }
 
+    private Set<Role> reorderAndAddDefaultRole(Set<Role> roles) {
```
scong-ttd, line 105: Is order important?

src/main/java/com/uid2/shared/auth/Roles.java
```diff
@@ -10,7 +10,9 @@
 
 public class Roles {
     public static <T extends Enum<T>> Set<T> getRoles(Class<T> type, JsonObject clientKeyJson) {
-        return getRoles(type, clientKeyJson.getJsonArray("roles", new JsonArray()));
+        return clientKeyJson.containsKey("roles")
+                ? getRoles(type, clientKeyJson.getJsonArray("roles"))
```
scong-ttd, line 14: Not critical here, but this scans the json object twice. If you are willing to break this one-liner you can avoid the additional pass:
```
JsonArray rolesArr = clientKeyJson.getJsonArray("roles");
return getRoles(type, rolesArr != null ? rolesArr : new JsonArray());
```

src/main/java/com/uid2/shared/auth/OperatorKey.java
sunnywu: i notice that when some other unrelated OperatorKey entries or fields are updated, the Roles ordering changed  such as this diff:

image

This could potentially cause us to think a diff has been created and therefore we need to write private site data again in UID-575. just FYI @aulme

scong-ttd: Makes sense! Thanks

PR #48, opened 2023-01-25
handle URI as CloudPath

scong-ttd:
Handle the case when an URL (potentially S3 path) works as a CloudPath

### DIFF
--- a/src/main/java/com/uid2/shared/store/CloudPath.java
+++ b/src/main/java/com/uid2/shared/store/CloudPath.java
@@ -1,35 +1,55 @@
 package com.uid2.shared.store;
 
+import java.net.URI;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


src/main/java/com/uid2/shared/store/CloudPath.java
```diff
@@ -38,7 +58,17 @@ public String toString() {
...
+        };
+
+        if (uriBase != null) {
+            return path.isEmpty() ?
+                    uriBase.getScheme() + "://" + uriBase.getAuthority() :
+                    uriBase.resolve("/").resolve(path).normalize().toString();
+        }
+
+        return path;
     }
```
sunnywu, line 73: should we use/compare uriBase in equals method and hashCode method below?

scong-ttd: nice catch, updated.

PR #49, opened 2023-01-26
Not listing files as part of getting metadata, instead providing tool…

aulme:
…s for Admin to do it on its own


scong-ttd:
LGTM

PR #51, opened 2023-01-30
Wrap attestation token lifetime into atService for better encapsulation

scong-ttd:
AttestationTokenService requires a `expiresAt` parameter when generating token, which should be kept internal because the caller (UIDCoreVerticle) shouldn't care about the lifetime.

### DIFF
--- a/src/main/java/com/uid2/shared/attest/AttestationTokenService.java
+++ b/src/main/java/com/uid2/shared/attest/AttestationTokenService.java
@@ -1,22 +1,42 @@
 package com.uid2.shared.attest;
 
 import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 
 public class AttestationTokenService implements IAttestationTokenService {
 
     private final String encryptionKey;
     private final String encryptionSalt;
+    private final long expiresAfterSeconds;
 
+    @Deprecated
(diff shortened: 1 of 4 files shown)


src/main/java/com/uid2/shared/attest/AttestationTokenService.java
```diff
@@ -1,22 +1,42 @@
...
 
+    public AttestationTokenService(String encryptionKey, String encryptionSalt, long expiresAfterSeconds) {
+        this.encryptionKey = encryptionKey;
+        this.encryptionSalt = encryptionSalt;
+        this.expiresAfterSeconds = expiresAfterSeconds;
+    }
+
+    @Override
+    public String createToken(String userToken) {
+        Instant expiresAt = Instant.now().plus(this.expiresAfterSeconds, ChronoUnit.SECONDS);
```
cYKatherine, line 27: We probably should do what `uid2-admin` was doing and create a wrapper around `Instant.now()`, but that's for future usage :)

cYKatherine (approved):

src/main/java/com/uid2/shared/attest/AttestationTokenService.java
scong-ttd: Thanks, let me keep this change small and upgrade everything in one go when we need to.

PR #52, opened 2023-02-01
Added a new method that allows to decide if no ACL return true or not

cody-constine-ttd:


src/main/java/com/uid2/shared/store/ACLSnapShotAccessMethod.java
```diff
@@ -0,0 +1,6 @@
+package com.uid2.shared.store;
+
+public enum ACLSnapShotAccessMethod {
```
jon8787, line 3: I got confused by this enum as ACL is used in both cases. Maybe something like:

AclMode.LegacyDep
AclMode.Sharing

?

jon8787 (dismissed):

src/main/java/com/uid2/shared/store/IKeysAclSnapshot.java
```diff
@@ -2,7 +2,10 @@
 
 import com.uid2.shared.auth.ClientKey;
 import com.uid2.shared.model.EncryptionKey;
+import com.uid2.shared.store.ACLMode.LegacyDEP;
 
 public interface IKeysAclSnapshot {
     boolean canClientAccessKey(ClientKey clientKey, EncryptionKey key);
```
mcollins-ttd, line 8: We should document how this overload behaves when there is no ACL.

src/main/java/com/uid2/shared/store/ACLMode/LegacyDEP.java
```diff
@@ -0,0 +1,6 @@
+package com.uid2.shared.store.ACLMode;
+
+public enum LegacyDEP {
+    noACLTrue,
```
mcollins-ttd, line 4: Suggestions:

```
public enum MissingAclMode {
  ALLOW_ALL,
  DENY_ALL
}
```

or

```
public enum NoAclMode {
  EMPTY_WHITELIST,
  EMPTY_BLACKLIST 
}
```

mcollins-ttd (approved):

PR #56, opened 2023-02-09
Upgraded Vert.x to 4.3.8

gmsdelmundo-ttd:


src/main/java/com/uid2/shared/model/EncryptionKey.java
```diff
@@ -69,6 +67,6 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        return Objects.hash(id, keyBytes, created, activates, expires, siteId);
+        return Objects.hash(id, Arrays.hashCode(keyBytes), created, activates, expires, siteId);
```
thomasm-ttd, line 70: Why did you change this? Not saying it is wrong, just want to understand

gmsdelmundo-ttd: `keyBytes` is of type `byte[]`, and `hashCode()` called directly on an array (i.e. `keyBytes.hashCode()`) will generate a hash based on the array's identity itself and not its contents

e.g.
```
arr1 = {1,2,3}, arr2 = {1,2,3}
arr1.hashCode() != arr2.hashCode() because arr1 is not arr2 (they are two distinct objects)
Arrays.hashCode(arr1) == Arrays.hashCode(arr2) because arr1's contents are identical to that of arr2
```

thomasm-ttd (approved):

PR #57, opened 2023-02-16
add tomestone flag bit into OptOutEntry to avoid major changes

scong-ttd:
Use 6th bit (from low) in the metadata byte to indicate user opt-in. This keeps all the data pipeline intact as we are not adding new record types. Only change required from consumer is to check that bit from the opt out record found.

### DIFF
--- a/src/main/java/com/uid2/shared/optout/OptOutEntry.java
+++ b/src/main/java/com/uid2/shared/optout/OptOutEntry.java
@@ -12,7 +12,8 @@ import java.util.Arrays;
 // Total: 72 bytes
 //
 // Metadata format:
-// Lower 6 bits  -- identity type
+// Lower 5 bits  -- identity type
+// Middle 1 bit  -- tombstone
 // Higher 2 bits -- record version
 // Total: 1 byte
 //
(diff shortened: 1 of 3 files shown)


src/main/java/com/uid2/shared/optout/OptOutEntry.java
```diff
@@ -131,16 +154,19 @@ public int hashCode() {
         return (int) (timestamp + Arrays.hashCode(identityHash) + Arrays.hashCode(advertisingId));
     }
 
-    private static byte calcMetadata(byte[] advertisingId) {
```
cody-constine-ttd, line 134: Do we need to leave the old function in for backward compatibility?

src/main/java/com/uid2/shared/optout/OptOutEntry.java
```diff
@@ -23,18 +24,33 @@ public final class OptOutEntry {
...
 
+    @Deprecated
     public OptOutEntry(byte[] identityHash, byte[] advertisingId, long ts) {
         assert identityHash.length == OptOutConst.Sha256Bytes;
         assert advertisingId.length == OptOutConst.Sha256Bytes || advertisingId.length == OptOutConst.Sha256Bytes + 1;
         // assert ts >= 0;
         this.identityHash = identityHash;
         this.advertisingId = advertisingId;
         this.timestamp = ts;
+        this.isTombstone = false;
```
atarassov-ttd, line 43: Can you call the constructor below instead?

PR #58, opened 2023-02-17
Randomize the attestation token expiry time

thomasm-ttd:
This will stop all operators expiring at the same time, causing the 401 alert to fire


src/test/java/com/uid2/shared/secure/AttestationTokenTest.java
```diff
@@ -67,6 +67,20 @@ public void testAttestationTokenNew() {
         Assertions.assertTrue(reconstructToken.getExpiresAt().isAfter(expiryLowerBound));
         Assertions.assertTrue(reconstructToken.getExpiresAt().isBefore(expiryUpperBound));
     }
+    @Test
+    public void testAttestationTokenNewDefaultLifetime() {
```
aulme, line 71: I think it's better if the test name describes what actual capability is being tested, "new default lifetime" is vague and won't be new any more soon.

src/test/java/com/uid2/shared/secure/AttestationTokenTest.java
```diff
@@ -67,6 +67,20 @@ public void testAttestationTokenNew() {
...
+        final long defaultLifetime = 7200;
+        final Instant expiryLowerBound = Instant.now().plusSeconds(defaultLifetime + (5 * 60));
+        final Instant expiryUpperBound = Instant.now().plusSeconds(defaultLifetime + (10 * 60));
+        final AttestationTokenService ats = new AttestationTokenService(ENCRYPTION_KEY, SALT);
+
+        final String attestationToken = ats.createToken("userToken");
+        Assertions.assertTrue(ats.validateToken("userToken", attestationToken));
+
+        final AttestationToken reconstructToken = AttestationToken.fromEncrypted(attestationToken, ENCRYPTION_KEY, SALT);
+        Assertions.assertTrue(reconstructToken.getExpiresAt().isAfter(expiryLowerBound));
```
aulme, line 81: We could make the tests more deterministic by constructor injecting a random number generator class and a clock into the AttestationTokenService instead of using statics. Then we could assert on exact equality of expected output rather than rely on bounds.

If you prefer to keep the bounds though, it's better to use AssertJ's `assertThat(reconstructToken.getExpiresAt()).isBefore(expiryUpperBound)` from `org.assertj.core.api.Assertions.assertThat` since this will give us a more useful error message when the test fails than `expected false  to be true`.

gmsdelmundo-ttd: For your latter comment, it's probably better to use `assertThat(reconstructToken.getExpiresAt()).isBetween(expiryLowerBound, expiryUpperBound)`

src/main/java/com/uid2/shared/attest/AttestationTokenService.java
```diff
@@ -24,7 +23,8 @@ public AttestationTokenService(String encryptionKey, String encryptionSalt, long
 
     @Override
     public String createToken(String userToken) {
-        Instant expiresAt = Instant.now().plus(this.expiresAfterSeconds, ChronoUnit.SECONDS);
+        long ran = ThreadLocalRandom.current().nextLong(300, 600); // random time between 5 and 10 minutes more to create some variation between when operators expire
```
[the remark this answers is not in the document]
thomasm-ttd: changed the name

src/test/java/com/uid2/shared/secure/AttestationTokenTest.java
thomasm-ttd: Updated the name of the test

src/test/java/com/uid2/shared/secure/AttestationTokenTest.java
thomasm-ttd: I agree, and have update the service as well as the tests. Left the constructors in there so we don't have to refactor Core and Optout at the same time.

PR #61, opened 2023-03-03
Fixed issue of operators not downloading all paritions

cody-constine-ttd:


scong-ttd:
LGTM. How can we easily tell if an operator is using permanent optout or only downloading 30 days of data?

scong-ttd (approved):

PR #71, opened 2023-05-14
uid2-1113 Adding the ability to set tags on objects in S3

thomasm-ttd:


src/main/java/com/uid2/shared/cloud/CloudStorageS3.java
```diff
@@ -173,18 +175,48 @@ public String mask(String cloudPath) {
         return cloudPath;
     }
 
+    @Override
+    public void setTags(String cloudPath, Map<String, String> tags) throws CloudStorageException {
```
vishalegbert-ttd, line 179: What calls setTags?

src/main/java/com/uid2/shared/cloud/TaggableCloudStorage.java
```diff
@@ -0,0 +1,8 @@
+package com.uid2.shared.cloud;
+
+import java.util.Map;
+
+public interface TaggableCloudStorage extends ICloudStorage {
+
+    void setTags(String cloudPath, Map<String, String> tags) throws CloudStorageException;
```
vishalegbert-ttd, line 7: Will we set object tags after they have been uploaded? Could we add a parameter for tags in the upload object method?

vishalegbert-ttd (approved):

src/main/java/com/uid2/shared/cloud/CloudStorageS3.java
```diff
@@ -51,6 +51,7 @@ public void upload(String localPath, String cloudPath) throws CloudStorageExcept
         try {
             File file = new File(localPath);
             this.s3.putObject(bucket, cloudPath, file);
+            this.checkVersioningEnabled();
```
mcollins-ttd, line 54: Can we instead check the value returned by `PutObjectResult.getVersionId()`?

https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/PutObjectResult.html#getVersionId--

PR #87, opened 2023-07-13
Support GCP OIDC token attestation

yishi-ttd:
### What does this PR do
- [GcpOidcAttestationProvider.java](https://github.com/IABTechLab/uid2-shared/compare/master...yishi-ttd:uid2-shared:support-gcp-oidc?expand=1#diff-f94cffef332a9734eb239e31d94521adc86f0f05ae5a9201544dcd4ca2fa502e) to handle new GCP attestation protocol `gcp-oidc`
- [TokenSignatureValidator.java](https://github.com/IABTechLab/uid2-shared/compare/master...yishi-ttd:uid2-shared:support-gcp-oidc?expand=1#diff-0887581f4ff014fd480ae020f411237e68fb402fa702c89a476dbaa19ed31e26) class to handle token validation & convertion logic

### Test Plan
UT

### TODO
- [PolicyValidator.java](https://github.com/IABTechLab/uid2-shared/compare/master...yishi-ttd:uid2-shared:support-gcp-oidc?expand=1#diff-1499f86832f80b57c9b0b8a2411c05e92a59ceb6d67606c70c7f19a222a69c26) implementation - after we closed the discussion how to handle env variable.
- hook it up in uid-core

### DIFF
--- a/src/main/java/com/uid2/shared/secure/GcpOidcAttestationProvider.java
+++ b/src/main/java/com/uid2/shared/secure/GcpOidcAttestationProvider.java
@@ -1,0 +1,95 @@
+package com.uid2.shared.secure;
+
+import com.uid2.shared.secure.gcpoidc.*;
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.nio.charset.StandardCharsets;
+import java.util.*;
+
+public class GcpOidcAttestationProvider implements IAttestationProvider{
+    private static final Logger LOGGER = LoggerFactory.getLogger(GcpOidcAttestationProvider.class);
+
+    private final ITokenSignatureValidator tokenSignatureValidator;
+
+    private final List<IPolicyValidator> supportedPolicyValidators;
+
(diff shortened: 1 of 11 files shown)


src/main/java/com/uid2/shared/secure/GcpOidcAttestationProvider.java
```diff
@@ -0,0 +1,95 @@
...
+    @Override
+    public void attest(byte[] attestationRequest, byte[] publicKey, Handler<AsyncResult<AttestationResult>> handler) {
+        try {
+            var tokenString = new String(attestationRequest, StandardCharsets.US_ASCII);
+
+            LOGGER.debug("Validating signature...");
+            var tokenPayload = tokenSignatureValidator.validate(tokenString);
+
+            if(this.Validate(tokenPayload)){
+                handler.handle(Future.succeededFuture(new AttestationResult(publicKey)));
```
vishalegbert-ttd, line 41: validation isn't async, is there any need to handle it asynchronously?

yishi-ttd: 1. Token signature validation:  involves API call to fetch GCP public key, but it will maintain a local cache. 
2. Token payload validation:  just some in-proc calculation. Should be super quick.

Let's keep it as is.

PR #88, opened 2023-07-17
Added a separate path to connect ot S3 via Iam rather than via user c…

aulme:
…redentials


src/main/java/com/uid2/shared/cloud/CloudStorageS3.java
```diff
@@ -46,6 +47,25 @@ public CloudStorageS3(String accessKeyId, String secretAccessKey, String region,
         this.bucket = bucket;
     }
 
+    public CloudStorageS3(String region, String bucket, String endpoint) {
```
thomasm-ttd, line 50: Is there a reason the parameter is called endpoint, but the other overload has one called s3Endpoint? Are they the same thing?

thomasm-ttd (approved):
One small comment, but happy to approve regardless

src/main/java/com/uid2/shared/cloud/CloudStorageS3.java
aulme: No, I just thought it's a better name, given the whole class is already about S3. Will change it for consistency though.

PR #89, opened 2023-07-18
OIDC payload check logic

yishi-ttd:
#### What does this MR do? 
- Add logic to validate OIDC payload, and generate enclave id.
- Refine test a bit and fix a typo bug in TokenParse logic.

#### Test Plan
UT

#### Where should the reviewer start?
`src/main/java/com/uid2/shared/secure/gcpoidc/PolicyValidator.java`


lunwang-ttd (approved):
LGTM

PR #90, opened 2023-07-19
Refactor UidCoreClient with separate AttestionTokenRetriever

cYKatherine:


src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -0,0 +1,195 @@
...
+        return cipher.doFinal(payload);
+    }
+
+    private static KeyPair generateKeyPair() throws NoSuchAlgorithmException {
+        KeyPairGenerator gen = KeyPairGenerator.getInstance(Const.Name.AsymetricEncryptionKeyClass);
+        gen.initialize(2048, new SecureRandom());
+        return gen.generateKeyPair();
+    }
+
+    public void notifyResponseStatusWatcher(int statusCode) {
```
vishalegbert-ttd, line 190: I think this whole notification thing should be reworked

- This method should be private
- We should create a new enum called AttestationResponseStatus: Success, InvalidKey, AttestationFailed, Unavailable
- When calling attest internal, we should notify the response watcher with the relevant enum
- We should update the disable watcher to only disable when the enum is InvalidKey or AttestationFailed

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -71,179 +75,88 @@ public ICloudStorage getContentStorage() {
...
-            attestInternal();
-
-        URLConnection conn = sendGet(url);
-
-        if (conn instanceof HttpURLConnection && attestIfRequired((HttpURLConnection) conn))
-            conn = sendGet(url);
-
-        return conn.getInputStream();
+    public boolean attested() {
+        return this.attestationToken.get() != null;
```
vishalegbert-ttd, line 83: What if it's expired?

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -0,0 +1,195 @@
...
+            throw new UidCoreClientException(ae);
+        } catch (IOException ioe) {
+            throw ioe;
+        } catch (Exception e) {
+            throw new UidCoreClientException(e);
+        }
+    }
+
+    public String getAttestationToken() { return this.attestationToken.get(); }
+    public void setAttestationToken(String atoken) {
```
vishalegbert-ttd, line 155: Make private

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -0,0 +1,195 @@
...
+        this.attestationToken.set(atoken);
+    }
+    public void setAttestationTokenExpiresAt(String expiresAt) {
```
vishalegbert-ttd, line 158: Make private

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -0,0 +1,195 @@
...
+            if (expiresAt == null) {
+                throw new UidCoreClientException(statusCode, "response json does not contain body.expiresAt");
+            }
+
+            atoken = new String(decrypt(Base64.getDecoder().decode(atoken), keyPair.getPrivate()), StandardCharsets.UTF_8);
+            LOGGER.info("Attestation successful. Attestation token received.");
+            setAttestationToken(atoken);
+            setAttestationTokenExpiresAt(expiresAt);
+
+            scheduleAttestationExpirationCheck();
```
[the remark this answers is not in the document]
vishalegbert-ttd: Only do this if there isn't a schedule

src/test/java/com/uid2/shared/attest/AttestationTokenRetrieverTest.java
```diff
@@ -0,0 +1,105 @@
...
+import java.net.URLConnection;
+import java.time.Instant;
+import java.util.concurrent.atomic.AtomicReference;
+import java.net.Proxy;
+import java.security.SecureRandom;
+
+import static com.uid2.shared.TestUtilites.makeInputStream;
+import static org.mockito.Mockito.*;
+
+public class AttestationTokenRetrieverTest {
```
vishalegbert-ttd, line 29: - Successful first attest, call attest(), expect call to httpClient, mock the response, check the token and expiry
- Attest again before expiry, move clock by 10 min, expect call to httpClient.............
- Test the response notification on success
- Test the response notification on failure
- Maybe test decryption stuff
-

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -0,0 +1,195 @@
...
+
+            if (statusCode < 200 || statusCode >= 300) {
+                LOGGER.warn("attestation failed with UID2 Core returning statusCode=" + statusCode);
+                throw new UidCoreClientException(statusCode, "unexpected status code from uid core service");
+            }
+
+            String responseBody = response.body();
+            JsonObject responseJson = (JsonObject) Json.decodeValue(responseBody);
+            if (isFailed(responseJson)) {
+                throw new UidCoreClientException(statusCode, "response did not return a successful status");
```
vishalegbert-ttd, line 127: Maybe this should AttestationTokenRetrieverException

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -1,64 +1,67 @@
...
+        this.appVersionHeader = appVersionHeader;
+    }
+
+    public UidCoreClient(String attestationEndpoint, String userToken, ApplicationVersion appVersion, Proxy proxy,
+                         IAttestationProvider attestationProvider, boolean enforceHttps, HttpClient httpClient) throws IOException {
+        this.proxy = proxy;
+        this.userToken = userToken;
+        this.contentStorage = new PreSignedURLStorage(proxy);
+        this.enforceHttps = enforceHttps;
+        this.responseWatcher = new AtomicReference<Handler<Integer>>(null);
```
aulme, line 61: Why an atomic reference? Is this going to be mutated by multiple threads?

src/test/java/com/uid2/shared/attest/AttestationTokenRetrieverTest.java
```diff
@@ -0,0 +1,104 @@
...
+        when(mockConn2.getOutputStream()).thenReturn(mockOutputStreamRequest);
+        when(mockConn2.getResponseCode()).thenReturn(200);
+        when(mockConn2.getInputStream()).thenReturn(new ByteArrayInputStream(content.toString().getBytes()));
+
+        URL mockURL = mock(URL.class);
+        when(mockURL.openConnection()).thenReturn(mockConn2);
+        when(mockURL.openConnection(any())).thenReturn(mockConn2);
+//        attestationTokenRetriever.setUrl(mockURL);
+
+        SecureRandom random = new SecureRandom(); // Default constructor uses a secure seed
```
aulme, line 93: How does this work? I don't see you passing this random anywhere

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -1,64 +1,67 @@
...
 
-    public static UidCoreClient createNoAttest(String attestationEndpoint, String userToken, ApplicationVersion appVersion, boolean enforceHttps) {
+    public static UidCoreClient createNoAttest(String attestationEndpoint, String userToken, ApplicationVersion appVersion, boolean enforceHttps) throws IOException {
         return new UidCoreClient(attestationEndpoint, userToken, appVersion, CloudUtils.defaultProxy, new NoAttestationProvider(), enforceHttps);
     }
 
     public UidCoreClient(String attestationEndpoint, String userToken, ApplicationVersion appVersion, Proxy proxy,
-                         IAttestationProvider attestationProvider, boolean enforceHttps) {
-        this.attestationEndpoint = attestationEndpoint;
+                         IAttestationProvider attestationProvider, boolean enforceHttps) throws IOException {
```
thomasm-ttd, line 39: It is generally better to make one constructor call the other. That way you reduce the duplication of code. In this case, the only difference is the httpClient, so in the constructor, check if the passed on httpClient is null, and if so, create a new one.

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -71,179 +74,94 @@ public ICloudStorage getContentStorage() {
...
-        return body.getString("attestation_token");
+    public void setResponseStatusWatcher(Handler<Integer> watcher) {
+        this.responseWatcher.set(watcher);
     }
 
-    private static byte[] decrypt(byte[] payload, PrivateKey privateKey) throws Exception {
-        Cipher cipher = Cipher.getInstance(Const.Name.AsymetricEncryptionCipherClass);
-        cipher.init(Cipher.DECRYPT_MODE, privateKey);
-        return cipher.doFinal(payload);
+    public void setHttpClient(HttpClient httpClient) {
```
thomasm-ttd, line 148: Why do we need this outside of the constructor?

cYKatherine: Fixed!

src/main/java/com/uid2/shared/attest/UidCoreClient.java
cYKatherine: Yes, in operator:
```
String clientsMdPath = this.config.getString(Const.Config.ClientsMetadataPathProp);
        this.clientKeyProvider = new RotatingClientKeyProvider(fsStores, new GlobalScope(new CloudPath(clientsMdPath)));
        String keysMdPath = this.config.getString(Const.Config.KeysMetadataPathProp);
        this.keyStore = new RotatingKeyStore(fsStores, new GlobalScope(new CloudPath(keysMdPath)));
        String keysAclMdPath = this.config.getString(Const.Config.KeysAclMetadataPathProp);
        this.keyAclProvider = new RotatingKeyAclProvider(fsStores, new GlobalScope(new CloudPath(keysAclMdPath)));
        String saltsMdPath = this.config.getString(Const.Config.SaltsMetadataPathProp);
        this.saltProvider = new RotatingSaltProvider(fsStores, saltsMdPath);
   ```
   It was called by 4 providers

Comment (cYKatherine) [reply src/main/java/com/uid2/shared/attest/UidCoreClient.java]: After talking with Thomas, I've removed `AtomicReference` for `responeWatcher` and handle the thread safety when actually calling `notifyResponseStatusWatcher`.

File: src/main/java/com/uid2/shared/IClock.java [hunk 12]
```diff
@@ -0,0 +1,9 @@
+package com.uid2.shared;
+
+import java.time.Instant;
+
+public interface IClock {
```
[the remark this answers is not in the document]
cYKatherine: Java Clock use `Clock.systemUTC().instant()).toMinutes()` to get the current time. It's a static method so we can't mock it in the test.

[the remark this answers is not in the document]
thomasm-ttd: Have a look at other places that use the clock - they generally take an instance of Clock on the constructor, and that way you can pass in a mock clock for testing. I like to provide an overload of the constructor that does require a clock being passed in, but then creates one before calling the next constructor.

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -0,0 +1,205 @@
...
+            catch (AttestationTokenRetrieverException | IOException e) {
+                notifyResponseStatusWatcher(401);
+                LOGGER.info("Re-attest failed: ", e.getMessage());
+            }
+        }
+    }
+
+    private void scheduleAttestationExpirationCheck() {
+        // Schedule the task to run every minute
+        executor.scheduleAtFixedRate(this::attestationExpirationCheck, 0, TimeUnit.MINUTES.toMillis(1), TimeUnit.MILLISECONDS);
```
thomasm-ttd, line 92: What happens if you try to call scheduleAtFixedRate if it is already scheduled?

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -0,0 +1,205 @@
...
+        } else {
+            this.executor = executor;
+        }
+    }
+
+    private void attestationExpirationCheck(){
+        Instant currentTime = clock.now();
+        Instant tenMinutesBeforeExpire = attestationTokenExpiresAt.minusSeconds(600);
+
+        stopAttestationExpirationCheck();
```
thomasm-ttd, line 76: If attestation fails for any reason, this will mean it has stopped and it will not reattest. I think this should turn off the scheduler, do the reattest attempt, and then turn on the scheduler again, even if attestation has failed.

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -0,0 +1,205 @@
...
+    private void scheduleAttestationExpirationCheck() {
+        // Schedule the task to run every minute
+        executor.scheduleAtFixedRate(this::attestationExpirationCheck, 0, TimeUnit.MINUTES.toMillis(1), TimeUnit.MILLISECONDS);
+    }
+
+    private void stopAttestationExpirationCheck() {
+        executor.shutdown();
+    }
+
+    public void attestInternal() throws IOException, AttestationTokenRetrieverException {
```
thomasm-ttd, line 99: This is not longer an internal method

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -0,0 +1,205 @@
...
+            }
+            requestJson.put("components", components);
+
+            HttpRequest httpRequest = HttpRequest.newBuilder()
+                    .setHeader("Content-Type", "application/json")
+                    .uri(URI.create(attestationEndpoint))
+                    .POST(HttpRequest.BodyPublishers.ofString(requestJson.toString(), StandardCharsets.UTF_8))
+                    .build();
+
+//            connection.setDoOutput(true);
```
thomasm-ttd, line 120: What are these comments for?

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -0,0 +1,205 @@
...
+        return responseJson.getString("status") == null || !responseJson.getString("status").equals("success");
+    }
+
+    private static KeyPair generateKeyPair() throws NoSuchAlgorithmException {
+        KeyPairGenerator gen = KeyPairGenerator.getInstance(Const.Name.AsymetricEncryptionKeyClass);
+        gen.initialize(2048, new SecureRandom());
+        return gen.generateKeyPair();
+    }
+
+    private void notifyResponseStatusWatcher(int statusCode) {
```
thomasm-ttd, line 191: You can leave it if you want, but you can move the null check outside the lock

src/main/java/com/uid2/shared/attest/AttestationTokenRetrieverException.java
```diff
@@ -0,0 +1,22 @@
...
+    public AttestationTokenRetrieverException(String message) {
+        super(message);
+    }
+
+    public AttestationTokenRetrieverException(int statusCode, String message) {
+        super("http status: " + String.valueOf(statusCode) + ", " + message);
+        this.statusCode = statusCode;
+    }
+
+    public boolean couldBeAttestationFailure() {
```
thomasm-ttd, line 19: should probably be called isAttestationFailure

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -1,64 +1,52 @@
...
     private boolean allowContentFromLocalFileSystem = false;
-    private AtomicReference<Handler<Integer>> responseWatcher;
+    private HttpClient httpClient;
+    private AttestationTokenRetriever attestationTokenRetriever;
+    private Handler<Integer> responseWatcher;
 
-    public static UidCoreClient createNoAttest(String attestationEndpoint, String userToken, ApplicationVersion appVersion, boolean enforceHttps) {
-        return new UidCoreClient(attestationEndpoint, userToken, appVersion, CloudUtils.defaultProxy, new NoAttestationProvider(), enforceHttps);
+    public static UidCoreClient createNoAttest(String attestationEndpoint, String userToken, ApplicationVersion appVersion, boolean enforceHttps) throws IOException {
+        return new UidCoreClient(attestationEndpoint, userToken, appVersion, CloudUtils.defaultProxy, new NoAttestationProvider(), enforceHttps, null);
```
thomasm-ttd, line 35: if, instead of passing null for httpClient, you passed HttpClient.newHttpClient(), you can get rid of the test in the other constructor.

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -1,64 +1,52 @@
...
         this.userToken = userToken;
-        this.appVersion = appVersion;
-        this.attestationProvider = attestationProvider;
         this.contentStorage = new PreSignedURLStorage(proxy);
-        this.attestationToken = new AtomicReference<>(null);
         this.enforceHttps = enforceHttps;
-        this.responseWatcher = new AtomicReference<Handler<Integer>>(null);
+        this.attestationTokenRetriever = new AttestationTokenRetriever(
+                attestationEndpoint, appVersion, attestationProvider, responseWatcher, new InstantClock(), null, null, null);
+        if (httpClient == null)
```
thomasm-ttd, line 46: please add {}

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -71,179 +59,90 @@ public ICloudStorage getContentStorage() {
...
-            try {
-                attestInternal();
-            }
-            catch (UidCoreClientException | IOException e) {
-                notifyResponseStatusWatcher(statusCode);
-                throw e;
-            }
-        } else {
-            notifyResponseStatusWatcher(statusCode);
+            attestationTokenRetriever.attestInternal();
```
thomasm-ttd, line 93: This code will probably not get executed unless attestation is failing as you have just called attestInternal above.

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -71,179 +59,90 @@ public ICloudStorage getContentStorage() {
...
-        connection.setRequestMethod(httpMethod);
-        connection.setRequestProperty(Const.Http.AppVersionHeader, appVersionHeader);
+        HttpRequest.Builder httpRequestBuilder = HttpRequest.newBuilder()
+                .uri(uri)
+                .GET()
+                .setHeader(Const.Http.AppVersionHeader, appVersionHeader);
 
         if(this.userToken != null && this.userToken.length() > 0) {
-            connection.setRequestProperty("Authorization", "Bearer " + this.userToken);
+            httpRequestBuilder.setHeader("Authorization", "Bearer " + this.userToken);
```
thomasm-ttd, line 113: Correct the spacing around the if( - should be if (

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -71,179 +59,90 @@ public ICloudStorage getContentStorage() {
...
         }
-
-        final String atoken = this.attestationToken.get();
-        if(atoken != null && atoken.length() > 0) {
-            connection.setRequestProperty("Attestation-Token", atoken);
+        if(attestationToken != null && attestationToken.length() > 0) {
```
thomasm-ttd, line 115: If the Bearer token or the attestation token are null or empty, there is no point making the call to Core as we know it will fail

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -71,179 +59,90 @@ public ICloudStorage getContentStorage() {
...
-        return cipher.doFinal(payload);
+    public void setResponseStatusWatcher(Handler<Integer> watcher) {
+        this.responseWatcher = watcher;
     }
 
-    private static KeyPair generateKeyPair() throws NoSuchAlgorithmException {
-        KeyPairGenerator gen = KeyPairGenerator.getInstance(Const.Name.AsymetricEncryptionKeyClass);
-        gen.initialize(2048, new SecureRandom());
-        return gen.generateKeyPair();
+    public void setAttestationTokenRetriever(AttestationTokenRetriever attestationTokenRetriever) {
```
thomasm-ttd, line 133: It's better to have the passed in on the constructor, rather than being able to change it.

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -71,179 +59,90 @@ public ICloudStorage getContentStorage() {
...
-        this.responseWatcher.set(watcher);
+    public void setAllowContentFromLocalFileSystem(boolean allowContentFromLocalFileSystem) {
+        this.allowContentFromLocalFileSystem = allowContentFromLocalFileSystem;
     }
 
-    public void notifyResponseStatusWatcher(int statusCode) {
-        Handler<Integer> w = this.responseWatcher.get();
-        if (w != null)
-            w.handle(statusCode);
+    public void setEnforceHttps(boolean enforceHttps) {
```
thomasm-ttd, line 145: Why don't you change it on the constructor?

src/test/java/com/uid2/shared/attest/AttestationTokenRetrieverTest.java
thomasm-ttd: You need some tests around the scheduling - making sure that before 10 min, you don't reattest, but after 10 min you do.

src/main/java/com/uid2/shared/attest/UidCoreClient.java
cYKatherine: Talked offline. We agreed on this `if (httpResponse.statusCode() == 401)` clause should never happen, but keeping this part of the code just to be extra safe. Adding a comment to it.

src/main/java/com/uid2/shared/attest/UidCoreClient.java
cYKatherine: Keeping this part to be consistent with the previous behaviour.

src/main/java/com/uid2/shared/IClock.java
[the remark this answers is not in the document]
cYKatherine: Yes. The other place where it passed the mock clock for testing are the customised Clock created by either @aulme or me. 

The places where it uses `Java.time.Clock` to get the current time uses `Clock.systemUTC().instant()).toMinutes()` which is a static method. We can't mock `systemUTC()` as we call it on the class. That's why this `IClock` interface here is needed.

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
cYKatherine: Added a check for that.

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -0,0 +1,203 @@
...
+import java.util.Base64;
+import java.util.Map;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+public class AttestationTokenRetriever {
+    private static final Logger LOGGER = LoggerFactory.getLogger(UidCoreClient.class);
```
thomasm-ttd, line 30: The class should be AttestationTokenRetriever.class

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -0,0 +1,203 @@
...
+        } catch (AttestationException ae) {
+            throw new AttestationTokenRetrieverException(ae);
+        } catch (IOException ioe) {
+            throw ioe;
+        } catch (Exception e) {
+            throw new AttestationTokenRetrieverException(e);
+        }
+    }
+
+    public String getAttestationToken() { return this.attestationToken.get(); }
```
thomasm-ttd, line 155: Can you match the { positions with the rest of the class - on a new line please

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -0,0 +1,203 @@
...
+        try {
+            if (this.responseWatcher != null)
+                this.responseWatcher.handle(statusCode);
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    public boolean attested() {
+        if (this.attestationToken.get() != null && this.clock.now().isBefore(this.attestationTokenExpiresAt)) return true;
```
thomasm-ttd, line 196: Can you put {} around the return true, and put on a new line

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -0,0 +1,203 @@
...
+    private void attestationExpirationCheck(long timerId) {
+        Instant currentTime = clock.now();
+        Instant tenMinutesBeforeExpire = attestationTokenExpiresAt.minusSeconds(600);
+
+        if (currentTime.isAfter(tenMinutesBeforeExpire)) {
+            LOGGER.info("Attestation token is 10 mins from the expiry timestamp %s. Re-attest...", attestationTokenExpiresAt);
+            long oldTimerId = this.timerId;
+            try {
+                attest();
+                this.vertx.cancelTimer(oldTimerId);
```
thomasm-ttd, line 79: What restarts the timer? The first time we attest, timerId == -1, so we set the timer. On the next attest, the timerId is not -1, so we don't restart it, and we never reset the timerId to -1, so I don't see how it gets scheduled again

cYKatherine: You are right!!! Fixed

aulme (approved):

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -0,0 +1,201 @@
...
+        }
+        if (attestationTokenDecryptor == null) {
+            this.attestationTokenDecryptor = new AttestationTokenDecryptor();
+        } else {
+            this.attestationTokenDecryptor = attestationTokenDecryptor;
+        }
+    }
+
+    private void attestationExpirationCheck(long timerId) {
+        Instant currentTime = clock.now();
```
thomasm-ttd, line 67: Now that you don't un-schedule it, what happens if the attest takes longer than 60sec? This will get called again, and will start another attest.
I suggest we add a flag around this code, so that if we are still in the process of attest, when this gets called again, it just exists without trying to attest again. We should also log that we have skipped the attest so we can detect if that is happening too much

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -0,0 +1,210 @@
...
+        if (!this.isExpiryCheckScheduled) {
+            // Schedule the task to run every minute
+            this.vertx.setPeriodic(0, 60000, this::attestationExpirationCheck);
+            this.isExpiryCheckScheduled = true;
+        }
+    }
+
+    public void attest() throws IOException, AttestationTokenRetrieverException {
+        try {
+            this.isAttesting = true;
```
thomasm-ttd, line 97: I suggest we move this to inside the attestationExpirationCheck, and put it just before the call to attest (line 78), and then in a finally block, set it to false.
As it is here, if attesting throws an error, the flag will never get reset to false, so it will never re-attest

thomasm-ttd (approved):

PR #91, opened 2023-07-20
allow by pass oidc signature check

yishi-ttd:
Allow bypass OIDC signature check (Do not check against GCP public cert) to faciliate local test.

It's not easy to get a Google signed JWT for local test - and we need to make sure it's not expired.


lunwang-ttd (approved):
LGTM

PR #97, opened 2023-07-28
Get JWT from attest response

thomasm-ttd:


src/main/java/com/uid2/shared/attest/JwtService.java
```diff
@@ -0,0 +1,142 @@
...
+
+    private final HashSet<PublicKey> publicKeys = new HashSet<>();
+
+
+    public JwtService(JsonObject config) {
+        this.config = config;
+        JsonArray publicKeys = config.getJsonArray(Const.Config.AwsKmsJwtSigningPublicKeysProp);
+        if (publicKeys == null || publicKeys.isEmpty()) {
+            LOGGER.error("Unable to read public keys from the configuration. JWTs can not be verified.");
+            return;
```
aulme, line 34: Can this recover? Should we throw an exception instead of keeping `publicKeys` null and then erroring at the consumption?

thomasm-ttd: There will be cases where JWT validation is not required, so the hosting service can set the config to not to enforce JWT validation. They would then not want to provide public keys. Also, the NullPointer exception for validateJwt would be caught and returned as an invalid JWT, rather than thrown.

aulme: Should we still log this as an error, then?

src/main/java/com/uid2/shared/secure/GcpOidcAttestationProvider.java
```diff
@@ -37,8 +37,9 @@ public void attest(byte[] attestationRequest, byte[] publicKey, Handler<AsyncRes
             LOGGER.debug("Validating signature...");
             var tokenPayload = tokenSignatureValidator.validate(tokenString);
 
+            //TODO: set the enclaveId once the registration of oidc enclaves is done
             if(this.Validate(tokenPayload)){
-                handler.handle(Future.succeededFuture(new AttestationResult(publicKey)));
+                handler.handle(Future.succeededFuture(new AttestationResult(publicKey, "TODO")));
```
aulme, line 42: Is this supposed to stay todo for now?

thomasm-ttd: Yes, as the GCP OIDC work is still in progress. I don't know what that will be yet

src/main/java/com/uid2/shared/attest/JwtService.java
thomasm-ttd: I'll change it to a info, as it may not be an error in that case

aulme (approved):

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -47,16 +50,12 @@ public UidCoreClient(String attestationEndpoint, String userToken, ApplicationVe
             this.httpClient = httpClient;
         }
         if (attestationTokenRetriever == null) {
-            this.attestationTokenRetriever = new AttestationTokenRetriever(
-                    attestationEndpoint, appVersion, attestationProvider, responseWatcher, new InstantClock(), null, null);
+            throw new IllegalArgumentException("AttestationTokenProvider can not be null");
```
thomasm-ttd, line 53: Fix this

src/test/java/com/uid2/shared/attest/JwtServiceTest.java
```diff
@@ -0,0 +1,137 @@
...
+        var validationResponse = service.validateJwt(EXPIRED_TOKEN, AUDIENCE, ISSUER);
+
+        assertNotNull(validationResponse);
+        assertFalse(validationResponse.getIsValid());
+        assertNotNull(validationResponse.getValidationException());
+        assertEquals("Token is expired", validationResponse.getValidationException().getMessage());
+    }
+
+    @Test
+    void throwsErrorIfNoKeyId() {
```
thomasm-ttd, line 101: Check this

PR #100, opened 2023-08-01
Restore AesCbc and AesGcm Helper Functions to Support Old and New Operator

zaiweidu:
We've updated AesCbc and AesGcm helper functions to support new Operator with KeySets. However, during the interim period, we still need old version of these APIs back to support old Operator with Keys and Key_Acls.

### DIFF
--- a/src/main/java/com/uid2/shared/encryption/AesCbc.java
+++ b/src/main/java/com/uid2/shared/encryption/AesCbc.java
@@ -16,11 +16,7 @@ public class AesCbc {
     private static final String cipherScheme = "AES/CBC/PKCS5Padding";
 
     public static EncryptedPayload encrypt(byte[] b, KeysetKey key) {
-        try {
-            return encrypt(b, key.getKeyBytes(), key.getKeyIdentifier());
-        } catch (Exception e) {
-            throw new RuntimeException("Unable to Encrypt", e);
-        }
+        return encrypt(b, key.getKeyBytes(), key.getKeyIdentifier());
     }
 
     private static EncryptedPayload encrypt(byte[] b, byte[] secretBytes, KeyIdentifier keyIdentifier) {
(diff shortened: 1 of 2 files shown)


thomasm-ttd (approved):

src/main/java/com/uid2/shared/encryption/AesCbc.java
```diff
@@ -54,4 +55,36 @@ public static byte[] decrypt(byte[] encryptedBytes, SecretKey key) {
             throw new RuntimeException("Unable to Decrypt", e);
         }
     }
+
+    // TODO: after KeySets fully migrated, below APIs shall be removed.
+    public static EncryptedPayload encrypt(byte[] b, EncryptionKey key) {
```
jon8787, line 60: is there a way to avoid the code duplication?

zaiweidu: Refactored the multiple encrypt and decrypt functions.

zaiweidu (commented):
Refactored multiple encrypt and decrypt functions in AesCbc and AesGcm.

src/main/java/com/uid2/shared/encryption/AesCbc.java
jon8787: Let's have a single encrypt with try/catch, and a single decrypt with try/catch.
The other overloads should be 1 line only

cody-constine-ttd (approved):
LGTM

src/main/java/com/uid2/shared/encryption/AesCbc.java
zaiweidu: Refactored encrypt/decrypt to single one with try/catch. Removed redundant try/catch.

zaiweidu (commented):
Ready for review.

jon8787 (approved):

PR #101, opened 2023-08-01
Added keyHash field for SHA-512-salted-hashed client and operator keys

gmsdelmundo:


src/main/java/com/uid2/shared/secret/SecureKeyGenerator.java
```diff
@@ -22,11 +28,57 @@ public String generateRandomKeyString(int keyLen) {
     }
 
     @Override
-    public String generateFormattedKeyString(int keyLen) {
+    public KeyGenerationResult generateFormattedKeyStringAndKeyHash(String keyPrefix, int keyLen) {
```
vishalegbert-ttd, line 31: There's too much going on here, the salting and hashing should take after the prefix has been prepended

Why does the hash need a prefix prepended?

Can we just feed the UTF-8 bytes of the full key string into the hash function?

src/main/java/com/uid2/shared/secret/SecureKeyGenerator.java
```diff
@@ -22,11 +28,57 @@ public String generateRandomKeyString(int keyLen) {
...
+
+        String hash = Utils.toBase64String(hashBytes); // This will always convert the hashBytes to a String with 88 chars (86 + 2 padding)
+        String salt = Utils.toBase64String(saltBytes);
+        String keyHash = String.format("%s$%s", hash, salt);
+
+        return new KeyGenerationResult(keyPrefix + formattedKey, keyPrefix + keyHash);
+    }
+
+    @Override
+    public boolean compareFormattedKeyStringAndKeyHash(String formattedInputKey, String keyHash) {
```
vishalegbert-ttd, line 50: This shouldn't exist either, let's keep the hashing as simple as possible. Mostly to support the old key format as well

src/main/java/com/uid2/shared/auth/ClientKey.java
```diff
@@ -25,8 +27,9 @@ public class ClientKey implements IRoleAuthorizable<Role> {
     private int siteId;
     private boolean disabled;
 
-    public ClientKey(String key, String secret, String name, String contact, Instant created, Set<Role> roles, int siteId, boolean disabled) {
+    public ClientKey(String key, String keyHash, String secret, String name, String contact, Instant created, Set<Role> roles, int siteId, boolean disabled) {
```
vishalegbert-ttd, line 30: I'm still sort of keen to have the salt as its own property.. 

Is there a strong need to do extra processing to parse out the salt from the same field?

src/test/java/com/uid2/shared/middleware/AttestationMiddlewareTest.java
```diff
@@ -65,8 +64,9 @@ void setUp() {
     @Test
     void trustedValidJwtNoRolesReturnsSuccess() throws JwtService.ValidationException {
         var attestationMiddleware = getAttestationMiddleware(true);
-        JwtValidationResponse response = new JwtValidationResponse(true).withSubject(hashedKey);
-        when(this.jwtService.validateJwt("dummy jwt", this.jwtAudience, this.jwtIssuer)).thenReturn(response);
+        JwtValidationResponse response = new JwtValidationResponse(true)
+                .withSubject(EXPECTED_OPERATOR_KEY_HASH);
```
vishalegbert-ttd, line 68: I'm guessing this test is passing because, as we learned, the subject validation is skipped?

vishalegbert-ttd (approved):

PR #107, opened 2023-08-12
UID2-882 site domain names

Ian-Nara:
[UID2-882](https://atlassian.thetradedesk.com/jira/browse/UID2-882)
add site_reader, dataclass, parser, and test


sunnywu:
looks ok to me. please don't merge it until the other admin PR is ready (ideally don't wanna have these Site class/helper duplicated in old code in admin and new shared at the same time

src/main/java/com/uid2/shared/model/Site.java
```diff
@@ -0,0 +1,59 @@
...
+        this.domainNames = domains;
+    }
+
+    public int getId() { return id; }
+    public String getName() { return name; }
+    public Boolean isEnabled() { return enabled; }
+    public Set<String> getDomainNames() { return domainNames; }
+
+    public void setEnabled(Boolean enabled) { this.enabled = enabled; }
+    public void setDomainNames(Set<String> domainNames) { this.domainNames = domainNames; }
```
mcollins-ttd, line 36: Do we need this? We're exposing `domainNames` through `getDomainNames`, so that might be enough for users of this class.

src/test/java/com/uid2/shared/store/RotatingSiteStoreTest.java
```diff
@@ -0,0 +1,102 @@
...
+import com.uid2.shared.model.Site;
+import com.uid2.shared.store.reader.RotatingClientSideKeypairStore;
+import com.uid2.shared.store.reader.RotatingSiteStore;
+import com.uid2.shared.store.scope.GlobalScope;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
```
mcollins-ttd, line 14: Unfortunately we have a mix of JUnit 4 and JUnit 5 tests. Please use JUnit 5 for new tests.

src/main/java/com/uid2/shared/model/Site.java
Ian-Nara: Write operation to domain_names in admin is to "set" rather than "add/remove". Simpler to replace reference with the new set than to empty existing set and repopulate with new set.

src/test/java/com/uid2/shared/store/RotatingSiteStoreTest.java
Ian-Nara: changed to junit5

mcollins-ttd (approved):

PR #110, opened 2023-08-17
Check readiness of attestion provider

lunwang-ttd:


src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -107,14 +107,18 @@ private void attestationExpirationCheck(long timerId) {
...
                 LOGGER.info("Attestation token is 10 mins from the expiry timestamp {}. Re-attest...", attestationTokenExpiresAt);
-                try {
-                    this.isAttesting = true;
-                    attest();
-                } catch (AttestationTokenRetrieverException | IOException e) {
-                    notifyResponseStatusWatcher(401);
-                    LOGGER.info("Re-attest failed: ", e);
-                } finally {
-                    this.isAttesting = false;
+                if (!attestationProvider.isReady()) {
```
yishi-ttd, line 110: I would prefer return early pattern, instead of if/else.
```
if(abc) return;
xyz;
```

Help us reduce nested level.

lunwang-ttd: Sure, tried to follow the same style, let me update

lunwang-ttd: Updated.

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -101,21 +101,26 @@ private void attestationExpirationCheck(long timerId) {
...
-                    LOGGER.info("Re-attest failed: ", e);
-                } finally {
-                    this.isAttesting = false;
-                }
+            return;
+        }
+
+        Instant currentTime = clock.now();
+        Instant tenMinutesBeforeExpire = attestationTokenExpiresAt.minusSeconds(600);
+        if (currentTime.isAfter(tenMinutesBeforeExpire)) {
```
yishi-ttd, line 109: NIT:
` if (!currentTime.isAfter(tenMinutesBeforeExpire)) return;`

This will make code even cleaner. Each check only need carry a small code block and one more level of nesting could be avoid.

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -101,21 +101,26 @@ private void attestationExpirationCheck(long timerId) {
...
+            LOGGER.info("Attestation token is 10 mins from the expiry timestamp {}. Re-attest...", attestationTokenExpiresAt);
+            if (!attestationProvider.isReady()) {
+                LOGGER.warn("Attestation provider is not ready. Skip re-attest.");
+                return;
+            }
+
+            try {
+                this.isAttesting = true;
```
yishi-ttd, line 117: BTW we may want to put this `this.isAttesting` in the very beginning after the `this.isAttesting` check, otherwise there would be multiple entry issue. (Though this should not be called from multiple thread.)

It's still not thread safe.. up to you whether want to use CompareAndSwap.

lunwang-ttd: Good catch. This may be called by the periodic timer from multiple threads, let me fix it.

lunwang-ttd: Updated.

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -99,24 +100,31 @@ public AttestationTokenRetriever(Vertx vertx,
...
-                try {
-                    this.isAttesting = true;
-                    attest();
-                } catch (AttestationTokenRetrieverException | IOException e) {
-                    notifyResponseStatusWatcher(401);
-                    LOGGER.info("Re-attest failed: ", e);
-                } finally {
-                    this.isAttesting = false;
-                }
+            if (!attestationProvider.isReady()) {
```
vishalegbert-ttd, line 117: Maybe we should check this in the attest function as well and throw an exception if it's not ready

Other consumers of the retriever will call attest directly, it shouldn't proceed if the provider is not ready

yishi-ttd (approved):

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
lunwang-ttd: If we also check isReady in the attest function, then isReady check will be executed twice.

I am thinking if we must check isReady before each attest, probably we should not add a new isReady method but throw a NotReady exception by getAttestationRequest method. Basically the option 2 in https://atlassian.thetradedesk.com/jira/browse/UID2-1632

vishalegbert-ttd: I don't see any issue with isReady being called twice

But I don't mind if we throw an exception instead. In the attestationExpirationCheck we can catch and log, allowing it to be checked again a minute later. Up to you

lunwang-ttd: GCP isReady involves file opening which is not light. But considering attestation calls are not frequent, calling twice is fine.

Let me check isReady in attest also.

lunwang-ttd: Updated, thanks

vishalegbert-ttd (approved):

PR #111, opened 2023-08-22
Add a property for storing cloud account ID in client key

asloobq:
This is just a draft to verify if approach is correct


asloobq:
Closing as not needed

PR #116, opened 2023-08-28
Ian UID2-1677 service and service links

Ian-Nara:


src/main/java/com/uid2/shared/store/parser/ServiceParser.java
```diff
@@ -0,0 +1,37 @@
...
+    @Override
+    public ParsingResult<Map<Integer, Service>> deserialize(InputStream inputStream) throws IOException {
+        JsonArray spec = Utils.toJsonArray(inputStream);
+        final HashMap<Integer, Service> serviceMap = new HashMap<>();
+        for (int i = 0; i < spec.size(); i++) {
+            JsonObject serviceSpec = spec.getJsonObject(i);
+            int serviceId = serviceSpec.getInteger("service_id");
+            int siteId = serviceSpec.getInteger("site_id");
+            String name = serviceSpec.getString("name");
+            JsonArray rolesSpec = serviceSpec.getJsonArray("roles");
```
thomasm-ttd, line 24: What happens if rolesSpec is null? Can it be?

Ian-Nara: I have added some logic into Service.java to enforce non-null. The jsonWriter used in uid2-admin will write empty set as [ ]

src/test/java/com/uid2/shared/store/RotatingServiceLinkStoreTest.java
```diff
@@ -0,0 +1,93 @@
...
+        metadata.put("service_links", serviceLinks);
+        return metadata;
+    }
+
+    private ServiceLink addServiceLink(JsonArray content, String linkId, int serviceId, int siteId, String name) {
+
+        ServiceLink l = new ServiceLink(linkId, serviceId, siteId, name);
+
+
+        JsonObject service = new JsonObject();
```
asloobq, line 52: You can use the `mapFrom()` method to create the JsonObject from the class object.
`JsonObject service = JsonObject.mapFrom(l);`
Just a suggestion :)

src/main/java/com/uid2/shared/store/parser/ServiceLinkParser.java
```diff
@@ -0,0 +1,33 @@
...
+        for (int i = 0; i < spec.size(); i++) {
+            JsonObject serviceLinkSpec = spec.getJsonObject(i);
+            String linkId = serviceLinkSpec.getString("link_id");
+            int serviceId = serviceLinkSpec.getInteger("service_id");
+            int siteId = serviceLinkSpec.getInteger("site_id");
+            String name = serviceLinkSpec.getString("name");
+
+            ServiceLink serviceLink = new ServiceLink(linkId, serviceId, siteId, name);
+
+            serviceLinkMap.put(linkId, serviceLink);
```
asloobq, line 28: ~Is the assumption that uniqueness of the `linkId` will be checked in the upper layer i.e. on receiving the POST request and its guaranteed to be unique when it gets here ?~
Hmm..link_id here doesn't have to be unique since its not generated by us. However, the combination of <link_id, service_id> should be unique. Does that fit into the existing code pattern ?

src/test/java/com/uid2/shared/store/RotatingServiceStoreTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+        final long count = serviceStore.loadContent(makeMetadata("locationPath"));
+        assertEquals(4, count);
+
+        assertEquals(s1, serviceStore.getService(1));
+        assertEquals(s2, serviceStore.getService(2));
+        assertEquals(s3, serviceStore.getService(3));
+        assertEquals(s4, serviceStore.getService(4));
+
+        assertTrue(serviceStore.getAllServices().contains(s1));
```
asloobq, line 94: nitpicking here: just use `containsAll()` ?
`assertTrue(serviceStore.getAllServices().containsAll(Arrays.asList(s1, s2, s3, s4)));`

asloobq (commented):
One important comment regarding link_id uniqueness. Others are minor

src/main/java/com/uid2/shared/store/IServiceLinkStore.java
```diff
@@ -0,0 +1,11 @@
+package com.uid2.shared.store;
+
+import com.uid2.shared.model.ServiceLink;
+
+import java.util.Collection;
+
+public interface IServiceLinkStore {
```
asloobq, line 7: I am wondering if we need another API here to fetch a record with a combination of <service_id, link_id>. In operator when a request comes in to /identity/map, the operator will validate the request by checking if the <service_id, link_id> combination exists. Would it be better to provide that API here or fetch all service links and then filter by service_id ? 
Your thoughts @Ian-Nara @thomasm-ttd

asloobq (commented):
One question about the API, otherwise LGTM

src/main/java/com/uid2/shared/store/IServiceLinkStore.java
Ian-Nara: I think it is okay to just get all and filter. I have implemented a similar thing for listing all links by site id, And this search by site id functionality also exists for the keys UI. If we want to update some fields in a service_link though we probably want this anyway.

src/main/java/com/uid2/shared/store/parser/ServiceParser.java
thomasm-ttd: If roles is not specified in the json, the call roleSpec.size below will throw an exception

thomasm-ttd (approved):

src/main/java/com/uid2/shared/store/parser/ServiceParser.java
Ian-Nara: Roles should be required to be specified in the json like the other fields, but we can allow it to be empty json array.

PR #117, opened 2023-08-29
Implement authorizable store

vishalegbert-ttd:


src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -0,0 +1,75 @@
...
+
+
+public class AuthorizableStore<T extends IAuthorizable> {
+    private class AuthorizableStoreSnapshot {
+        private final Map<ByteBuffer, T> authorizables;
+        private final Collection<byte[]> salts;
+
+        public AuthorizableStoreSnapshot(Collection<T> authorizables) {
+            this.authorizables = authorizables.stream()
+                    .filter(a -> a.getKeyHash() != null) // TODO: remove this filter when all keys have hashes
```
gmsdelmundo, line 18: Admins, clients and operators should now all have key hash, so this should be good to remove

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -0,0 +1,75 @@
...
+                    .collect(Collectors.toMap(
+                            a -> ByteBuffer.wrap(Base64.getDecoder().decode(a.getKeyHash())),
+                            a -> a
+                    ));
+            this.salts = authorizables.stream()
+                    .filter(a -> a.getKeySalt() != null) // TODO: remove this filter when all keys have salts
```
gmsdelmundo, line 24: Admins, clients and operators should now all have key salt, so this should be good to remove

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -0,0 +1,75 @@
...
+            return authorizables;
+        }
+
+        public Collection<byte[]> getSalts() {
+            return salts;
+        }
+    }
+
+    private final AtomicReference<AuthorizableStoreSnapshot> authorizables;
+    private final KeyHasher keyHasher;
```
gmsdelmundo, line 39: `keyHasher` can be a static final variable

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public AuthorizableStore() {
+        this.authorizables = new AtomicReference<>(new AuthorizableStoreSnapshot(new ArrayList<>()));
+        this.keyHasher = new KeyHasher();
+    }
+
+    public void refresh(Collection<T> authorizables) {
+        this.authorizables.set(new AuthorizableStoreSnapshot(authorizables));
```
gmsdelmundo, line 47: No need for `this.`

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.nio.ByteBuffer;
+import java.util.*;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
+
+
+public class AuthorizableStore<T extends IAuthorizable> {
+    private class AuthorizableStoreSnapshot {
+        private final Map<ByteBuffer, T> authorizables;
```
gmsdelmundo, line 13: Can we use `byte[]` instead as the key?

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public T getFromKeyHash(String keyHash) {
+        AuthorizableStoreSnapshot latest = authorizables.get();
+
+        byte[] keyHashBytes;
+        try {
+            keyHashBytes = Base64.getDecoder().decode(keyHash);
+        } catch (IllegalArgumentException e) {
+            return null;
```
gmsdelmundo, line 70: Should we log out the exception here?

src/main/java/com/uid2/shared/store/parser/ClientParser.java
```diff
@@ -10,6 +10,7 @@
 import java.util.HashMap;
 import java.util.Map;
 
+// change to return simple list
```
gmsdelmundo, line 13: What is this referring to?

src/main/java/com/uid2/shared/store/reader/RotatingClientKeyProvider.java
```diff
@@ -39,8 +40,10 @@
 */
 public class RotatingClientKeyProvider implements IClientKeyProvider, StoreReader<Collection<ClientKey>> {
     private final ScopedStoreReader<Map<String, ClientKey>> reader;
+    private final AuthorizableStore<ClientKey> authorizableStore;
```
gmsdelmundo, line 43: Add a newline below this line

src/main/java/com/uid2/shared/store/reader/RotatingClientKeyProvider.java
```diff
@@ -55,14 +58,20 @@ public long getVersion(JsonObject metadata) {
 
     @Override
     public long loadContent(JsonObject metadata) throws Exception {
-        return reader.loadContent(metadata, "client_keys");
+        long version = reader.loadContent(metadata, "client_keys");
+        this.authorizableStore.refresh(this.getAll());
```
gmsdelmundo, line 62: Are we now essentially maintaining two different "snapshots" that are duplicated? `AuthorizableStore` 's records are loaded in by `reader`'s snapshot, would we still need `reader`'s snapshot?

src/main/java/com/uid2/shared/store/reader/RotatingClientKeyProvider.java
```diff
@@ -55,14 +58,20 @@ public long getVersion(JsonObject metadata) {
...
+        return version;
     }
 
     @Override
     public ClientKey getClientKey(String token) {
         return reader.getSnapshot().get(token);
     }
 
+    public ClientKey getClientKeyFromHash(String hash) {
```
gmsdelmundo, line 71: Should we update the `IClientKeyProvider` interface with the above method?

src/test/java/com/uid2/shared/auth/AuthorizableStoreTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+        KeyHashResult clientHash1 = keyHasher.hashKey("test-key-1");
+        clients.add(new ClientKey("", clientHash1.getHash(), clientHash1.getSalt(), "", "client1"));
+
+        KeyHashResult clientHash2 = keyHasher.hashKey("test-key-2");
+        clients.add(new ClientKey("", clientHash2.getHash(), clientHash2.getSalt(), "", "client2"));
+
+        KeyHashResult clientHash3 = keyHasher.hashKey("test-key-3");
+        clients.add(new ClientKey("", clientHash3.getHash(), clientHash3.getSalt(), "", "client3"));
+
```
gmsdelmundo, line 30: There's an extra newline here

src/test/java/com/uid2/shared/auth/AuthorizableStoreTest.java
```diff
@@ -0,0 +1,79 @@
+package com.uid2.shared.auth;
+
+import com.uid2.shared.secret.KeyHashResult;
+import com.uid2.shared.secret.KeyHasher;
+import org.junit.Assert;
```
gmsdelmundo, line 5: We should be using JUnit Jupiter's asserts

`import static org.junit.jupiter.api.Assertions.*;`

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
vishalegbert-ttd: Removed

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
vishalegbert-ttd: Removed

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
vishalegbert-ttd: Yeah possibly, I left it as non static incase we decide to inject it later on, but if you disagree I can make it static

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
vishalegbert-ttd: Removed

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
vishalegbert-ttd: byte[] doesn't support proper hashcode, ByteBuffer does apparently

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
vishalegbert-ttd: Added log

src/main/java/com/uid2/shared/store/parser/ClientParser.java
vishalegbert-ttd: This is a note for you :) 

In this parser, we are building a dictionary from keys to ClientKey, this can be changed to return a list of ClientKey instead

Removed the comment

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
gmsdelmundo: One thing we could do instead is take the `keyHasher` in as a parameter instead in the constructor so we can pass the same `keyHasher` instance from our main Verticle code to all the different providers

src/main/java/com/uid2/shared/store/reader/RotatingClientKeyProvider.java
vishalegbert-ttd: ScopedReader maintains its own snapshot of what its parsed, which might be necessary for other users for ScopedReader

It's only the references that are duplicated, I don't see an issue with that. That's the cost of abstraction sometimes

src/main/java/com/uid2/shared/store/reader/RotatingClientKeyProvider.java
vishalegbert-ttd: Yep, done

src/test/java/com/uid2/shared/auth/AuthorizableStoreTest.java
vishalegbert-ttd: Removed

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
vishalegbert-ttd: All the Stores will need a keyHasher for almost no reason, I'd rather make it static then

src/main/java/com/uid2/shared/store/reader/RotatingClientKeyProvider.java
gmsdelmundo: I see, I brought up the question because `RotatingOperatorKeyProvider` does it a bit differently without `ScopedReader` - maybe it would be investigating whether it would be good to make this consistent on the operator key side

vishalegbert-ttd: Ah right, I think this scoped reader stuff is for site specific files, I don't think Operators are site specific so they didn't need it 

For Operator I guess you can just pass in a collection of Operators directly, the RotatingOperatorKeyProvider won't need to store a duplicate list of references

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
gmsdelmundo: I was referring to making it as a parameter only for `RotatingClientKeyProvider`, `RotatingOperatorKeyProvider`, `AdminUserProvider` since those are the classes that would need to make use of the `keyHasher`. We won't have to do it for others

src/main/java/com/uid2/shared/store/reader/RotatingClientKeyProvider.java
gmsdelmundo: Makes sense

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -0,0 +1,78 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.nio.ByteBuffer;
+import java.util.*;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
+
+
+public class AuthorizableStore<T extends IAuthorizable> {
+    private class AuthorizableStoreSnapshot {
```
gmsdelmundo, line 14: We can make this static as the nested class doesn't reference anything from the outer class

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
gmsdelmundo: Right, makes sense

I think as long as we don't modify the keys directly (which we shouldn't be doing anyway), this should be fine

pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>com.uid2</groupId>
     <artifactId>uid2-shared</artifactId>
-    <version>4.9.0-3481c8dd0a</version>
+    <version>4.9.4-SNAPSHOT</version>
```
gmsdelmundo, line 9: Don't forget to revert this back

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
vishalegbert-ttd: I just tried it, but it complained it couldn't access the generic type I think

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
vishalegbert-ttd: I just made it static 😅

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
gmsdelmundo: Ah right... because it does reference something with the other class, which is the `T`

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -0,0 +1,77 @@
...
+        public Map<ByteBuffer, T> getAuthorizables() {
+            return authorizables;
+        }
+
+        public Collection<byte[]> getSalts() {
+            return salts;
+        }
+    }
+
+    private static final KeyHasher keyHasher = new KeyHasher();
```
gmsdelmundo, line 38: Could you rename this to `KEY_HASHER`?

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -0,0 +1,77 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(AuthorizableStore.class);
```
gmsdelmundo, line 39: Can this be renamed to `LOGGER`?

gmsdelmundo (commented):
Last two comments

pom.xml
vishalegbert-ttd: Done

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
vishalegbert-ttd: Done

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
vishalegbert-ttd: Done

gmsdelmundo (approved):

PR #119, opened 2023-09-07
Added caching to authorizable store and updated client key provider

gmsdelmundo:


src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -1,77 +1,154 @@
...
         }
         return null;
     }
 
-    public T getFromKeyHash(String keyHash) {
-        AuthorizableStoreSnapshot latest = authorizables.get();
+    private static LoadingCache<String, String> createCache() {
+        return Caffeine.newBuilder()
+                .maximumSize(CACHE_MAX_SIZE)
+                .recordStats()
```
aulme, line 94: No way to get to those stats for now - until we export them, let's not calculate them.

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -1,77 +1,154 @@
...
+            return Base64.getDecoder().decode(str);
         } catch (IllegalArgumentException e) {
-            LOGGER.error("Invalid base64 key hash: {}", keyHash);
+            LOGGER.error("Invalid base64 string: {}", str);
+            return null;
+        }
+    }
+
+    private static Integer getSiteIdFromKey(String key) {
+        Pattern keyPattern = Pattern.compile("(?:UID2|EUID)-[CO]-[LTIP]-([0-9]+)-.{6}\\..{38}");
```
aulme, line 113: Move to static so it doesn't need to be re-compiled

src/test/java/com/uid2/shared/auth/ClientKeyTest.java
```diff
@@ -1,61 +1,61 @@
...
-import io.vertx.core.json.JsonObject;
-import org.junit.Assert;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.junit.Test;
 
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
 public class ClientKeyTest {
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
aulme, line 10: Let's move ObjectMapper creation to factory so we can apply common settings

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -1,77 +1,153 @@
...
+        Integer siteId = getSiteIdFromKey(key);
+
+        if (siteId != null) {
+            T authorizable = snapshot.getAuthorizableBySiteId(siteId);
+            if (authorizable == null) {
+                return null;
+            }
+
+            byte[] hash = KEY_HASHER.hashKey(key, convertBase64StringToBytes(authorizable.getKeySalt()));
+            return snapshot.getAuthorizableByHash(ByteBuffer.wrap(hash));
```
vishalegbert-ttd, line 75: You already have the authorizable, why not check hash equality here?

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -1,77 +1,153 @@
...
+
+        AuthorizableStoreSnapshot latest = authorizables.get();
+        return latest.getAuthorizableByHash(hashBytes);
+    }
+
+    private T getAuthorizableBySiteId(AuthorizableStoreSnapshot snapshot, String key) {
+        Integer siteId = getSiteIdFromKey(key);
+
+        if (siteId != null) {
+            T authorizable = snapshot.getAuthorizableBySiteId(siteId);
```
vishalegbert-ttd, line 69: Can there be multiple authorizables per site ID? I think so

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
vishalegbert-ttd: Doesn't apply if there are multiple salts to consider per site ID

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -1,77 +1,153 @@
...
+            }
+
+            byte[] hash = KEY_HASHER.hashKey(key, convertBase64StringToBytes(authorizable.getKeySalt()));
+            return snapshot.getAuthorizableByHash(ByteBuffer.wrap(hash));
+        }
+
+        return null;
+    }
+
+    private T getAuthorizableIfKeyExists(AuthorizableStoreSnapshot snapshot, String key) {
```
vishalegbert-ttd, line 81: Rename to getAuthorizableByKey?

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -1,77 +1,153 @@
...
             return null;
         }
+    }
+
+    private class AuthorizableStoreSnapshot {
+        private final Map<ByteBuffer, T> hashToAuthorizableMap;
+        private final Map<Integer, T> siteIdToAuthorizableMap;
+        private final Set<byte[]> salts;
+
+        public AuthorizableStoreSnapshot(Collection<T> hashToAuthorizableMap) {
```
vishalegbert-ttd, line 126: Why is this called hashToAuthorizableMap?

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -1,77 +1,153 @@
...
+                            a -> wrapHashToByteBuffer(a.getKeyHash()),
+                            a -> a));
+            this.siteIdToAuthorizableMap = hashToAuthorizableMap.stream()
+                    .filter(a -> a.getSiteId() != null)
+                    .collect(Collectors.toMap(
+                            IAuthorizable::getSiteId,
+                            a -> a));
+            this.salts = hashToAuthorizableMap.stream()
+                    .map(a -> convertBase64StringToBytes(a.getKeySalt()))
+                    .collect(Collectors.toSet());
```
vishalegbert-ttd, line 138: Why set?

src/main/java/com/uid2/shared/auth/ClientKey.java
```diff
@@ -110,6 +113,7 @@ public String getName() {
         return name;
     }
 
+    @Deprecated // N/A
```
vishalegbert-ttd, line 116: Why are these deprecated?

src/main/java/com/uid2/shared/store/parser/ClientParser.java
```diff
@@ -1,25 +1,25 @@
...
-        JsonArray keysSpec = Utils.toJsonArray(inputStream);
-        Map<String, ClientKey> keyMap = new HashMap<>();
-        for (int i = 0; i < keysSpec.size(); ++i) {
-            JsonObject keySpec = keysSpec.getJsonObject(i);
-            ClientKey clientKey = ClientKey.valueOf(keySpec);
-            keyMap.put(clientKey.getKey(), clientKey);
-        }
-        return new ParsingResult<>(keyMap, keysSpec.size());
+        ClientKey[] clientKeys = OBJECT_MAPPER.readValue(inputStream, ClientKey[].class);
+        Map<String, ClientKey> keyMap = Stream.of(clientKeys).collect(Collectors.toMap(
```
vishalegbert-ttd, line 19: Do we need to build this map anymore? Can it be a simple list instead now?

src/test/java/com/uid2/shared/auth/AuthorizableStoreTest.java
```diff
@@ -5,74 +5,102 @@
...
 public class AuthorizableStoreTest {
-    private AuthorizableStore<ClientKey> store;
+    private static final KeyHasher KEY_HASHER = new KeyHasher();
+    private static final Instant NOW = Instant.now();
+    private static final String CLIENT_KEY_1 = "UID2-C-L-11-abcdef.abcdefabcdefabcdefabcdefabcdefabcdefab";
+    private static final String CLIENT_KEY_2 = "UID2-C-L-12-ghijkl.ghijklghijklghijklghijklghijklghijklgh";
+    private static final String CLIENT_KEY_3 = "UID2-C-L-13-mnopqr.mnopqrmnopqrmnopqrmnopqrmnopqrmnopqrmn";
+    private static final String CLIENT_KEY_4 = "UID2-C-L-14-stuvwx.stuvwxstuvwxstuvwxstuvwxstuvwxstuvwxst";
+
+    private static final String ADMIN_KEY_1 = "UID2-A-L-abcdef.abcdefabcdefabcdefabcdefabcdefabcdefab";
```
vishalegbert-ttd, line 22: Are these used?

src/test/java/com/uid2/shared/auth/AuthorizableStoreTest.java
```diff
@@ -5,74 +5,102 @@
...
+import static org.junit.jupiter.api.Assertions.*;
 
 public class AuthorizableStoreTest {
-    private AuthorizableStore<ClientKey> store;
+    private static final KeyHasher KEY_HASHER = new KeyHasher();
+    private static final Instant NOW = Instant.now();
+    private static final String CLIENT_KEY_1 = "UID2-C-L-11-abcdef.abcdefabcdefabcdefabcdefabcdefabcdefab";
+    private static final String CLIENT_KEY_2 = "UID2-C-L-12-ghijkl.ghijklghijklghijklghijklghijklghijklgh";
+    private static final String CLIENT_KEY_3 = "UID2-C-L-13-mnopqr.mnopqrmnopqrmnopqrmnopqrmnopqrmnopqrmn";
+    private static final String CLIENT_KEY_4 = "UID2-C-L-14-stuvwx.stuvwxstuvwxstuvwxstuvwxstuvwxstuvwxst";
```
vishalegbert-ttd, line 20: Maybe add a non site ID key just to test those code paths?

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
vishalegbert-ttd: Should probably add support for it and maybe these should be cached. Perhaps the site ID only reduces the number of salts to search, might make the code cleaner

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
gmsdelmundo: I've updated the method's functionality so maybe a different name?

src/main/java/com/uid2/shared/auth/ClientKey.java
gmsdelmundo: I've marked it that way because it doesn't make sense to have constructors that sets a default site ID

The only times we've used those constructors is under three situations.

**Situation 1 - Never (N/A)**

**Situation 2 - Test cases only**
It would be better to create a builder that provides more meaningful defaults. Otherwise, we would have production code that are actually only used for testing, and also makes it a bit more confusing (e.g. why are we defaulting site ID to 0 or -1?).

**Situation 3 - Builder pattern where it's not really needed**
There are places in our code, specifically uid2-admin, where we use a constructor that has default values, but use the `.with()` functions to provider values for those default-initialized fields anyway.

src/main/java/com/uid2/shared/store/parser/ClientParser.java
gmsdelmundo: I don't think so - its output is used as:
`ParsingResult<T> parsed = parser.deserialize(inputStream);`

And this `T` is part of `ScopedStoreReader<T>`

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
vishalegbert-ttd: Other functions that return nulls don't have the If_Exists suffix, I'd prefer less verbose names but up to you

src/main/java/com/uid2/shared/store/parser/ClientParser.java
vishalegbert-ttd: We can change the T to List<ClientKeys>?

src/test/java/com/uid2/shared/auth/AuthorizableStoreTest.java
```diff
@@ -5,74 +5,181 @@
...
+    }
 
-        this.store.refresh(List.of(client4));
+    private ClientKey createClientKey(KeyHashResult khr, String name, int siteId) {
+        return new ClientKey("", khr.getHash(), khr.getSalt(), "", name, NOW, Set.of(), siteId);
+    }
 
-        assertEquals("client4", this.store.getFromKey("test-key-4").getName());
-        assertNull(this.store.getFromKey("test-key-1"));
+    private static class SitelessAuthorizable implements IAuthorizable {
```
vishalegbert-ttd, line 129: Why do we need this? Can't we just have a ClientKey with a legacy key?

src/main/java/com/uid2/shared/auth/AuthorizableStore.java
```diff
@@ -1,77 +1,137 @@
...
 
     public void refresh(Collection<T> authorizablesToRefresh) {
         authorizables.set(new AuthorizableStoreSnapshot(authorizablesToRefresh));
     }
 
-    public T getFromKey(String key) {
+    public T getAuthorizableByKey(String key) {
+        AuthorizableStoreSnapshot latest = authorizables.get();
+
+        String cachedHash = keyToHashCache.getIfPresent(key);
```
vishalegbert-ttd, line 37: Can we add a simple prometheus counter to count cache misses?

src/test/java/com/uid2/shared/auth/AuthorizableStoreTest.java
gmsdelmundo: This was to basically test AdminUser-type authorizables as well, or I guess this can be replaced with a null-site ID client key?

vishalegbert-ttd: Yeah whatever works, I more so meant we should use a plaintext key without a site ID in some test cases. The actual class doesn't matter much

gmsdelmundo: Yep, I've added two tests:
1. The one that's essentially the legacy client key (`getAuthorizableByKey_returnsClientKey_withLegacyFormat`)
2. The one that emulates a null-site ID authorizable like AdminUser (`getAuthorizableByKey_returnsKey_withNoSiteId`)

vishalegbert-ttd (approved):

PR #121, opened 2023-09-11
disallow EU prefix region

yishi-ttd:
Check below payload example for the structure:

https://github.com/IABTechLab/uid2-shared/blob/master/src/test/resources/com.uid2.shared/test/secure/gcpoidc/jwt_payload_full.json

GCP zones: https://cloud.google.com/compute/docs/regions-zones


lunwang-ttd (approved):
LGTM

PR #125, opened 2023-09-12
Ian UI d2 1394 update attest failure reaction

Ian-Nara:
Update shared to inform attestation response watcher of both status code and response body.


src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -104,20 +104,14 @@ private InputStream readContentFromLocalFileSystem(String path, Proxy proxy) thr
...
+    private InputStream get(String path, String jwtToken) throws IOException, InterruptedException {
         String attestationToken = attestationTokenRetriever.getAttestationToken();
 
         HttpResponse<String> httpResponse;
         httpResponse = sendHttpRequest(path, attestationToken, jwtToken);
 
-        // This should never happen, but keeping this part of the code just to be extra safe.
         if (httpResponse.statusCode() == 401) {
             LOGGER.info("Initial response from UID2 Core returned 401, performing attestation");
-            attestationTokenRetriever.attest();
```
cody-constine-ttd, line 120: Why did you remove this?

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -33,12 +34,11 @@ public class AttestationTokenRetriever {
     private final AtomicReference<String> attestationToken;
     private final AtomicReference<String> optOutJwt;
     private final AtomicReference<String> coreJwt;
-    private final Handler<Integer> responseWatcher;
+    private final Handler<Pair<Integer, String>> responseWatcher;
```
cody-constine-ttd, line 37: Are you sure this is only used by operator?

src/main/java/com/uid2/shared/attest/UidCoreClient.java
Ian-Nara: We want to decouple the attest() calls from the calls to a refresh endpoint.

"The renewal of the attestation process should be changed to be done as a background process, rather than being done as part of a request - response call from a client of the Operator. This would allow for the retries to not block an incoming request."

Calls to attest are now call handled within the AttestationTokenRetreiver class

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
Ian-Nara: Out of the core, optout, admin, and operator, the attestation token retriever is used by operator and optout, but optout uses a null responseWatcher. Will look into optout closer to make sure it is not broken by these changes

zaiweidu (commented):
LGTM

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
Ian-Nara: I have tested optout/core with my shared changes locally and it is able to attest and refresh with no extra changes needed

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -96,16 +96,28 @@ public AttestationTokenRetriever(Vertx vertx,
         }
         this.appVersionHeader = builder.toString();
 
+        this.vertx.setPeriodic(0, attestCheckMilliseconds, this::attestationCheck);
```
thomasm-ttd, line 99: Why have you moved this to the constructor? Currently it is only set after the first attestation, but now it will happen 60s after the construction, regardless of whether attest has been called.

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -96,16 +96,28 @@ public AttestationTokenRetriever(Vertx vertx,
...
             return;
         }
 
         try {
+
+            if (!attestationProvider.isReady()) {
+                throw new AttestationTokenRetrieverException("attestation provider is not ready");
+            }
+
+            if(this.attestationToken.get() == null) {
```
thomasm-ttd, line 115: Why have you added this check and then attest? The scheduled attestationCheck should only happen after the first successful attestation

src/main/java/com/uid2/shared/attest/UidCoreClient.java
thomasm-ttd: This work was already done, and attestation now happens 10 minutes before the expiry of the token. I don't believe we need to change this. getWithAttest allows the first request to perform attestation, at which point the schedule will start.

src/test/java/com/uid2/shared/attest/AttestationTokenRetrieverTest.java
```diff
@@ -65,17 +64,42 @@ public void attest_succeed_attestationTokenSet(Vertx vertx, VertxTestContext tes
...
+        when(mockHttpResponse.body()).thenReturn("bad" ,"bad", "bad", expectedResponseBody);
+
+        when(mockHttpClient.send(any(HttpRequest.class), any(HttpResponse.BodyHandler.class))).thenReturn(mockHttpResponse);
+        when(mockAttestationTokenDecryptor.decrypt(any(), any())).thenReturn("test_attestation_token".getBytes(StandardCharsets.UTF_8));
+
+        when(clock.now()).thenReturn(Instant.parse("2023-08-01T00:00:00.111Z"));
+
         attestationTokenRetriever = getAttestationTokenRetriever(vertx);
+        testContext.awaitCompletion(3500, TimeUnit.MILLISECONDS);
+        verify(this.responseWatcher, times(3)).handle(Pair.of(401, "bad"));
```
thomasm-ttd, line 93: Why do we have 3 401 responses with 'bad' body and 3 401 responses  with the more detailed response?

src/test/java/com/uid2/shared/attest/AttestationTokenRetrieverTest.java
```diff
@@ -65,17 +64,42 @@ public void attest_succeed_attestationTokenSet(Vertx vertx, VertxTestContext tes
...
         Assertions.assertEquals("test_attestation_token", attestationTokenRetriever.getAttestationToken());
         Assertions.assertEquals("appName=appVersion;Component1=Value1;Component2=Value2", attestationTokenRetriever.getAppVersionHeader());
-        verify(this.responseWatcher, times(1)).handle(200);
+        verify(this.responseWatcher, times(1)).handle(Pair.of(200, expectedResponseBody));
         testContext.completeNow();
     }
 
     @Test
-    public void attest_currentTimeAfterTenMinsBeforeAttestationTokenExpiry_expiryCheckCallsAttest(Vertx vertx, VertxTestContext testContext) throws Exception {
+    public void attest_fixedIntervalUntilSuccess(Vertx vertx, VertxTestContext testContext) throws Exception {
```
thomasm-ttd, line 76: Good test

src/test/java/com/uid2/shared/attest/AttestationTokenRetrieverTest.java
```diff
@@ -65,17 +64,42 @@ public void attest_succeed_attestationTokenSet(Vertx vertx, VertxTestContext tes
...
+        String expectedResponseBody = "{\"body\": {\"attestation_token\": \"test\",\"expiresAt\": \"2023-08-03T09:09:30.608597Z\",\"attestation_jwt_optout\": \"\",\"attestation_jwt_core\": \"\"},\"status\": \"success\"}";
+        when(mockHttpResponse.statusCode()).thenReturn(401, 401, 401, 200);
+        when(mockHttpResponse.body()).thenReturn("bad" ,"bad", "bad", expectedResponseBody);
+
+        when(mockHttpClient.send(any(HttpRequest.class), any(HttpResponse.BodyHandler.class))).thenReturn(mockHttpResponse);
+        when(mockAttestationTokenDecryptor.decrypt(any(), any())).thenReturn("test_attestation_token".getBytes(StandardCharsets.UTF_8));
+
+        when(clock.now()).thenReturn(Instant.parse("2023-08-01T00:00:00.111Z"));
+
         attestationTokenRetriever = getAttestationTokenRetriever(vertx);
```
thomasm-ttd, line 91: maybe we should reduce the retry interval for the test. This test will take 3 seconds every time

thomasm-ttd (commented):
I don't believe we need to change the attestation scheduled behaviour. Attestation happens on the first call to core, and then after that 10 minutes before expiry of the token.

src/main/java/com/uid2/shared/attest/UidCoreClient.java
Ian-Nara: Ah I see I was not considering that the expiration check will handle the retry already for failure in all cases except if the first attestation call fails. I have undone this change.

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
Ian-Nara: changed back

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
Ian-Nara: changed back as noted above.

src/test/java/com/uid2/shared/attest/AttestationTokenRetrieverTest.java
Ian-Nara: reduced the interval

src/test/java/com/uid2/shared/attest/AttestationTokenRetrieverTest.java
Ian-Nara: modified this test. It now has an intial sucess followed by 2 failures, followed by a sucess

thomasm-ttd (approved):

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -112,6 +115,7 @@ private void attestationExpirationCheck(long timerId) {
         }
 
         try {
+
```
thomasm-ttd, line 118: Extra line

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -127,7 +131,7 @@ private void attestationExpirationCheck(long timerId) {
 
             attest();
         } catch (AttestationTokenRetrieverException | IOException e) {
```
thomasm-ttd, line 133: I know this was there before, but what would cause the IOException? Should we be returning a 401 for an IOException? Or is it a 500?

Ian-Nara: httpClient.post() is what throws it. Changed this to return a 500

Ian-Nara: Changed this to a 500 for an IOException, and fixed an issue in URLConnectionHttpClient.java that resulted in IOException being thrown for any non-200 status code

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
```diff
@@ -130,9 +129,12 @@ private void attestationExpirationCheck(long timerId) {
             }
 
             attest();
-        } catch (AttestationTokenRetrieverException | IOException e) {
+        } catch (AttestationTokenRetrieverException e) {
             notifyResponseWatcher(401, e.getMessage());
             LOGGER.info("Re-attest failed: ", e);
+        } catch (IOException e) {
```
thomasm-ttd: Sorry to do this as I know you have been trying to get this approved for a while, but can we log out for both the 401 and the 500 the response code, and change the messages so we can tell what the cause was?

src/test/java/com/uid2/shared/attest/AttestationTokenRetrieverTest.java
```diff
@@ -73,40 +74,35 @@ public void attest_succeed_attestationTokenSet(Vertx vertx, VertxTestContext tes
...
-        String expectedResponseBody = "{\"body\": {\"attestation_token\": \"test\",\"expiresAt\": \"2023-08-01T00:00:00.111Z\",\"attestation_jwt_optout\": \"\", \"attestation_jwt_core\": \"\"},\"status\": \"success\"}";
-        when(mockHttpResponse.body()).thenReturn(expectedResponseBody);
-        when(mockHttpResponse.statusCode()).thenReturn(200);
-
-        HttpResponse<String> mockHttpResponseSecondAttest = mock(HttpResponse.class);
-        String expectedResponseBodySecondAttest = "{\"body\": {\"attestation_token\": \"test\",\"expiresAt\": \"2023-08-01T00:00:00.111Z\",\"attestation_jwt_optout\": \"\",\"attestation_jwt_core\": \"\"},\"status\": \"success\"}";
-        when(mockHttpResponseSecondAttest.body()).thenReturn(expectedResponseBodySecondAttest);
-        when(mockHttpResponseSecondAttest.statusCode()).thenReturn(300);
+        String expectedResponseBody = "{\"body\": {\"attestation_token\": \"test\",\"expiresAt\": \"2023-08-01T00:00:00.111Z\",\"attestation_jwt_optout\": \"\",\"attestation_jwt_core\": \"\"},\"status\": \"success\"}";
+        when(mockHttpResponse.statusCode()).thenReturn(200, 401, 401, 200);
```
thomasm-ttd, line 90: We should use 500 as the error code here. If we use 401, it implies the operator will retry on a 401, which it won't, it will shutdown

thomasm-ttd (commented):
This will break the build for operator, so please make sure you are ready to merge operator soon after merging this change, publishing it and updating operator

src/main/java/com/uid2/shared/attest/AttestationTokenRetriever.java
Ian-Nara: I am not quite sure what you are suggesting. It will already log out for both a 401 and a 500, with the message being based on the exception that occurred. Are you suggesting we generalize the error that is logged, so that a 401 always logs the same thing and a 500 always logs the same thing? And if so what do you suggest for the message?

thomasm-ttd: What I want to make sure is that we can tell the difference in our logs (so we can alert differently) between the 2 types of 401 errors. If you are confident that this is the case, then I am OK with this.

thomasm-ttd (approved):
Please check that all the conversations have been resolved before merging

PR #127, opened 2023-09-13
Change UID2-Shared for OptOut Policy Check

zaiweidu:
1. Added a new method `getOldestClientKey()` to RotatingClientKeyProvider.java
2. Added unit tests to RotatingClientKeyProviderTest.java

### DIFF
--- a/src/main/java/com/uid2/shared/store/reader/RotatingClientKeyProvider.java
+++ b/src/main/java/com/uid2/shared/store/reader/RotatingClientKeyProvider.java
@@ -5,14 +5,14 @@ import com.uid2.shared.auth.ClientKey;
 import com.uid2.shared.auth.IAuthorizable;
 import com.uid2.shared.cloud.DownloadCloudStorage;
 import com.uid2.shared.store.CloudPath;
-import com.uid2.shared.store.ScopedStoreReader;
 import com.uid2.shared.store.IClientKeyProvider;
+import com.uid2.shared.store.ScopedStoreReader;
 import com.uid2.shared.store.parser.ClientParser;
 import com.uid2.shared.store.scope.StoreScope;
 import io.vertx.core.json.JsonObject;
 
 import java.util.Collection;
-import java.util.List;
+import java.util.Comparator;
 
 /*
   1. metadata.json format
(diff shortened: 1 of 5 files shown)


src/main/java/com/uid2/shared/store/reader/RotatingClientKeyProvider.java
```diff
@@ -93,4 +95,12 @@ public CloudPath getMetadataPath() {
     public IAuthorizable get(String key) {
         return getClientKey(key);
     }
+
+    public ClientKey getOldestClientKey(int siteId) {
+        List<ClientKey> keys = reader.getSnapshot().values().stream()
+                .filter(k -> k.getSiteId() == siteId) // filter by site id
+                .sorted(Comparator.comparing(ClientKey::getCreated)) // sort by key creation timestamp ascending
+                .collect(Collectors.toList());
```
atarassov-ttd, line 103: Would `findFirst()` be more appropriate here?

src/main/java/com/uid2/shared/store/reader/RotatingClientKeyProvider.java
```diff
@@ -93,4 +95,12 @@ public CloudPath getMetadataPath() {
     public IAuthorizable get(String key) {
         return getClientKey(key);
     }
+
+    public ClientKey getOldestClientKey(int siteId) {
```
atarassov-ttd, line 99: Should this be a method on the snapshot instead? This is not really specific to the rotating provider.

src/test/java/com/uid2/shared/auth/RotatingClientKeyProviderTest.java
```diff
@@ -0,0 +1,32 @@
...
+import com.uid2.shared.store.scope.GlobalScope;
+import io.vertx.core.json.JsonObject;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.*;
+
+public class RotatingClientKeyProviderTest {
+    @Test
+    public void testGetOldestClientKeyBySiteId() throws Exception {
```
atarassov-ttd, line 15: A test for when site does not have any keys?

src/main/java/com/uid2/shared/store/reader/RotatingClientKeyProvider.java
zaiweidu: Replaced `get(0)` with `findFirst()` as suggested.

src/main/java/com/uid2/shared/store/reader/RotatingClientKeyProvider.java
zaiweidu: Moved `getOldestClientKey()` to _ClientKeySnapshot_.

src/test/java/com/uid2/shared/auth/RotatingClientKeyProviderTest.java
zaiweidu: Test case added as suggested.

atarassov-ttd (approved):

PR #131, opened 2023-09-18
expose metrics for record counts (ScopedStoreReader)

jon8787:


src/main/java/com/uid2/shared/store/ScopedStoreReader.java
```diff
@@ -34,6 +38,11 @@ public ScopedStoreReader(DownloadCloudStorage fileStreamProvider, StoreScope sco
             this.contentStreamProvider = fileStreamProvider;
         }
         latestSnapshot = new AtomicReference<>();
+
+        Gauge.builder("uid2_config_store_entry_count", latestEntryCount::get)
+                .tag("store", dataTypeName)
+                .description("gauge for " + dataTypeName + "store total entry count")
```
atarassov-ttd, line 44: space before "store"

src/main/java/com/uid2/shared/store/ScopedStoreReader.java
```diff
@@ -55,8 +64,11 @@ private long loadContent(String path) throws Exception {
         try (InputStream inputStream = this.contentStreamProvider.download(path)) {
             ParsingResult<T> parsed = parser.deserialize(inputStream);
             latestSnapshot.set(parsed.getData());
-            LOGGER.info(String.format("Loaded %d %s", parsed.getCount(), dataTypeName));
-            return parsed.getCount();
+
+            final int count = parsed.getCount();
+            latestEntryCount.set(count);
```
atarassov-ttd, line 69: What happens in case of errors? If we don't have an active snapshot, would we be left with a previous count that could mask an issue?

src/main/java/com/uid2/shared/store/ScopedStoreReader.java
jon8787: fixed here and a few other places

src/main/java/com/uid2/shared/store/ScopedStoreReader.java
jon8787: I see RotatingStoreVerticle has `consecutive_refresh_failures`. But what if we had a metric "last_successful_refresh" and an alert when it's too old?

jon8787: setting to -1 on failure

jon8787: Decided to leave this for now. `getSnapshot()` will return whatever the last successful `loadContent()` retrieved, so the metric matches that. We can do refresh failure count as another ticket.

atarassov-ttd (approved):

PR #140, opened 2023-09-26
Azure Confidential Container MAA token parse and policy validation logic

yishi-ttd:
#### What does this MR do?
- Logic to parse MAA token, validate signature against MAA server, and check policy.
- Policy to check for MAA payload:
   - Is running in expected SKU (VM is AMD & Host Utility VM).
   - VM is not in debug mode.
   - Location is not China or EU.
   - PK matches in runtime info.

#### Test Plan
- UT to verify token parse and policy check logic
- E2E verified locally that the real token string can be parsed and validated by MAA

#### Where should the reviewer start
- `src/main/java/com/uid2/shared/secure/azurecc/AzurePublicKeyProvider.java` logic to provide Azure public key.
- `src/main/java/com/uid2/shared/secure/azurecc/MaaTokenSignatureValidator.java` logic to parse the token and validate signature against MAA
- `src/main/java/com/uid2/shared/secure/azurecc/PolicyValidator.java` logic to check policy against the parsed MAA payload.

#### Notes
- Why implements a new publickey provider？
  - MAA certs are stored as x5c(X.509 certificate chain), not supported by Google auth lib.
- How to generate `x-ms-runtime` field
  - When generate MAA token, we could pass a RunTime Info string to MAA sever. It's the base64 string of JSON payload: `{
    "location": "East US",
    "publicKey": "abc"
  }`


src/main/java/com/uid2/shared/secure/azurecc/MaaTokenPayload.java
```diff
@@ -0,0 +1,26 @@
...
+@Value
+@Builder(toBuilder = true)
+public class MaaTokenPayload {
+    public static final String SEV_SNP_VM_TYPE = "sevsnpvm";
+    public static final String AZURE_COMPLIANT_UVM = "azure-compliant-uvm";
+
+    private String attestationType;
+    private String complianceStatus;
+    private boolean vmDebuggable;
+    private String ccePolicy;
```
lunwang-ttd, line 15: Actually it's "hostData", should we rename it as "hostData", otherwise it's a bit confusing because in the deployment template "ccePolicy" means the actual policy instead of policy digest

lunwang-ttd: Or "ccePolicyDigest"?

src/main/java/com/uid2/shared/secure/azurecc/AzurePublicKeyProvider.java
```diff
@@ -0,0 +1,77 @@
...
+
+        var signers = client.listAttestationSigners().getAttestationSigners();
+
+        ImmutableMap.Builder<String, PublicKey> keyCacheBuilder = new ImmutableMap.Builder();
+
+        for (var signer : signers){
+            var keyId = signer.getKeyId();
+            var certs = signer.getCertificates();
+            if(!certs.isEmpty()){
+                var publicKey = certs.get(0).getPublicKey();
```
lunwang-ttd, line 65: Is there any chance that a certs contains more than 1 public key, and the second key matches with the public key in JWT?

yishi-ttd: Good question.
It's possible that there's a cert chain. But here we only care about the first level cert. We don't care about it's parent cert.
See below for example.
image

I could add some comments to elaborate in code.

src/main/java/com/uid2/shared/secure/azurecc/MaaTokenPayload.java
yishi-ttd: Will change it to ccePolicyDigest.

src/main/java/com/uid2/shared/secure/azurecc/AzurePublicKeyProvider.java
lunwang-ttd: Got it, thanks for the explanation

lunwang-ttd (approved):
LGTM

PR #150, opened 2023-10-06
Added an unknown default to Roles

cody-constine-ttd:


src/main/java/com/uid2/shared/store/parser/ClientParser.java
```diff
@@ -12,6 +13,10 @@
 public class ClientParser implements Parser<Collection<ClientKey>> {
     private static final ObjectMapper OBJECT_MAPPER = Mapper.getInstance();
 
+    public ClientParser() {
+        OBJECT_MAPPER.enable(DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE);
```
asloobq, line 17: I think putting this a `static` block is probably better since OBJECT_MAPPER is static.

asloobq (approved):

PR #151, opened 2023-10-10
Azure CC attestation provider

lunwang-ttd:


src/main/java/com/uid2/shared/Const.java
```diff
@@ -72,7 +72,9 @@ public static class Config {
         public static final String OptOutUrlProp = "optout_url";
         public static final String AwsKmsJwtSigningPublicKeysProp = "aws_kms_jwt_signing_public_keys";
         public static final String EnforceJwtProp = "enforceJwt";
+
         public static final String MaaServerBaseUrl="maa_server_base_url";
+        public static final String DefaultMaaServerBaseUrl="https://sharedeus.eus.attest.azure.net";
```
yishi-ttd, line 77: This should not be in this config section.
This section only contains config key.

DefaultMaaServerBaseUrl -> probably we not need this , we could enfore it be set in (operator/core) config.

lunwang-ttd: Agree that it's not consistent with the current practice.
But I think we need a place to define default config values, I would prefer define config key and its default value together. So that we can avoid hardcode default value like https://github.com/IABTechLab/uid2-core/blob/7f5063e1c2296b4f505c9e1ebba201f0b5486f88/src/main/java/com/uid2/core/Main.java#L122-L123

I can remove the default value for now, and will use the default value defined in uid2-attestation-azure after we figure out a better way to define default value.

lunwang-ttd: I removed the default value, and renamed it to `MaaServerBaseUrlProp`, will update core and operator accordingly

yishi-ttd (approved):

PR #157, opened 2023-11-03
Add description and visible flag to sites

alex-yau-ttd:
- Add `description` and `visible` to `Site`, including setters, getters and method overrides
- Update `SiteParser`
- Update tests

See https://github.com/IABTechLab/uid2-admin/pull/173 for relevant changes in admin

### DIFF
--- a/src/main/java/com/uid2/shared/model/Site.java
+++ b/src/main/java/com/uid2/shared/model/Site.java
@@ -29,7 +29,7 @@ public class Site {
                 @JsonProperty("description") String description,
                 @JsonProperty("enabled") Boolean enabled,
                 @JsonProperty("clientTypes") Set<ClientType> types,
-                @JsonProperty("domains") Set<String> domains,
+                @JsonProperty("domain_names") Set<String> domains,
                 @JsonProperty("visible") Boolean visible,
                 @JsonProperty("created") long created) {
         this.id = id;
--- a/src/main/java/com/uid2/shared/model/Site.java
+++ b/src/main/java/com/uid2/shared/model/Site.java
@@ -11,11 +11,13 @@ import java.util.Set;
 public class Site {
     private final int id;
     private final String name;
+    private String description;
     private Boolean enabled;
     @JsonProperty("domain_names")
     private Set<String> domainNames;
 
     private Set<ClientType> clientTypes;
+    private Boolean visible;
     private final long created;
 
     public Site(int id, String name, Boolean enabled) {
@@ -54,25 +56,46 @@ public class Site {
         this.created = created;
     }
 
+    public Site(int id, String name, String description, Boolean enabled, Set<ClientType> types, Set<String> domains, Boolean visible, long created) {
+        this.id = id;
+        this.name = name;
+        this.description = description;
+        this.enabled = enabled;
+        this.clientTypes = types;
+        this.domainNames = domains;
+        this.visible = visible;
+        this.created = created;
+    }
+
     public int getId() { return id; }
     public String getName() { return name; }
+    public String getDescription() { return description; }
     public Boolean isEnabled() { return enabled; }
     public Set<String> getDomainNames() { return domainNames; }
     public Set<ClientType> getClientTypes() { return clientTypes; }
+    public Boolean isVisible() { return visible; }
+
     public long getCreated() { return  created; }
+
+    public void setDescription(String description) { this.description = description; }
+
     public void setClientTypes(Set<ClientType> clientTypes) { this.clientTypes = clientTypes; }
 
     public void setEnabled(Boolean enabled) { this.enabled = enabled; }
     public void setDomainNames(Set<String> domainNames) { this.domainNames = domainNames; }
+    public void setVisible(Boolean visible) { this.visible = visible; }
+
 
     @Override
     public String toString() {
         return "Site{" +
                 "id=" + id +
                 ", name='" + name + '\'' +
+                ", description=" + description +
                 ", enabled=" + enabled +
                 ", domain_names=" + domainNames.toString() +
                 ", clientTypes=" + clientTypes.toString() +
+                ", visible=" + visible +
                 ", created=" + String.valueOf(created) +
                 '}';
     }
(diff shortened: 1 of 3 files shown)


src/main/java/com/uid2/shared/model/Site.java
```diff
@@ -54,25 +56,46 @@ public Site(int id, String name, Boolean enabled, Set<ClientType> types, Set<Str
...
+        this.enabled = enabled;
+        this.clientTypes = types;
+        this.domainNames = domains;
+        this.visible = visible;
+        this.created = created;
+    }
+
     public int getId() { return id; }
     public String getName() { return name; }
+    public String getDescription() { return description; }
```
alex-yau-ttd, line 72: I'm wondering if this should have a default of empty string - and how this would be implemented

alex-yau-ttd: actually, I am inclined to just leave the description to default to null given this does not seem to be a common pattern we use. besides, this field won't have any functional use.

src/main/java/com/uid2/shared/store/parser/SiteParser.java
```diff
@@ -23,7 +23,9 @@ public ParsingResult<Map<Integer, Site>> deserialize(InputStream inputStream) th
             JsonObject siteSpec = sitesSpec.getJsonObject(i);
             final int siteId = siteSpec.getInteger("id");
             final String name = siteSpec.getString("name");
+            final String description = siteSpec.getString("description");
             final Boolean enabled = siteSpec.getBoolean("enabled", false);
+            final Boolean visible = siteSpec.getBoolean("visible", true);
```
alex-yau-ttd, line 28: I think `true` is a reasonable default. Only internal sites and special case sites should not be `visible`

src/main/java/com/uid2/shared/model/Site.java
```diff
@@ -82,11 +105,11 @@ public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         Site site = (Site) o;
-        return id == site.id && name.equals(site.name) && enabled.equals(site.enabled) && domainNames.equals(site.domainNames) && clientTypes.equals(site.clientTypes) && created == site.created;
+        return id == site.id && name.equals(site.name) && description.equals(site.description) && enabled.equals(site.enabled) && domainNames.equals(site.domainNames) && clientTypes.equals(site.clientTypes) && visible.equals(site.visible) && created == site.created;
```
gmsdelmundo, line 108: Since `visible` and `description` will be null for existing objects, it may potentially raise a NPE here. Maybe better to check for null value for visible and description?

src/main/java/com/uid2/shared/store/parser/SiteParser.java
gmsdelmundo: Should we use ObjectMapper here instead? So that there's no need to manually update this parser each time, and we can have the default value of visible in the class

src/test/java/com/uid2/shared/store/RotatingSiteStoreTest.java
```diff
@@ -58,8 +58,10 @@ private Site addSite(JsonArray content, int siteId, String name, boolean enabled
         JsonObject site = new JsonObject();
```
gmsdelmundo, line 58: Same comment here about `ObjectMapper`

src/main/java/com/uid2/shared/model/Site.java
gmsdelmundo: Sounds good

src/main/java/com/uid2/shared/store/parser/SiteParser.java
gmsdelmundo: Yep let's use `true`

src/main/java/com/uid2/shared/model/Site.java
gmsdelmundo: Possibly no need for `visible` null check if we default it

src/main/java/com/uid2/shared/model/Site.java
alex-yau-ttd: flipped on this, I think I'll default to `""`

src/main/java/com/uid2/shared/model/Site.java
```diff
@@ -82,11 +134,11 @@ public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         Site site = (Site) o;
-        return id == site.id && name.equals(site.name) && enabled.equals(site.enabled) && domainNames.equals(site.domainNames) && clientTypes.equals(site.clientTypes) && created == site.created;
+        return id == site.id && name.equals(site.name) && Objects.equals(description, site.description) && enabled.equals(site.enabled) && domainNames.equals(site.domainNames) && clientTypes.equals(site.clientTypes) && visible.equals(site.visible) && created == site.created;
```
gmsdelmundo, line 137: I guess we don't need the Objects.equals here anymore then since we're defaulting to ""?

src/main/java/com/uid2/shared/store/parser/SiteParser.java
```diff
@@ -1,45 +1,24 @@
...
 import java.io.InputStream;
-import java.util.HashMap;
-import java.util.HashSet;
+import java.util.Arrays;
 import java.util.Map;
-import java.util.Set;
 import java.util.stream.Collectors;
 
-public class SiteParser implements Parser<Map<Integer, Site>> {
+public class SiteParser implements Parser<Map<Integer,Site>> {
```
gmsdelmundo, line 13: Missing space after the , so it registered as a change

src/main/java/com/uid2/shared/store/parser/SiteParser.java
```diff
@@ -1,45 +1,24 @@
...
+
```
gmsdelmundo, line 14: No need for this newline

src/test/java/com/uid2/shared/store/RotatingSiteStoreTest.java
alex-yau-ttd: I'm missing some understanding - how can we use `ObjectMapper` here? I think we can avoid using this `JsonObject()` and the numerous `site.put`s below - is this still using `OBJECT_MAPPER.readValue`?

src/main/java/com/uid2/shared/model/Site.java
alex-yau-ttd: yep good point - done

src/main/java/com/uid2/shared/store/parser/SiteParser.java
alex-yau-ttd: oopsie, fixed

src/main/java/com/uid2/shared/store/parser/SiteParser.java
alex-yau-ttd: removed

src/test/java/com/uid2/shared/store/RotatingSiteStoreTest.java
alex-yau-ttd: how's this?
```
private Site addSite(JsonArray content, int siteId, String name, String description, boolean enabled, boolean visible, long created, String... domains) throws IOException {

    Site s = new Site(siteId, name, description, enabled, new HashSet<>(), new HashSet<>(List.of(domains)), visible, created);

    String jsonString = OBJECT_MAPPER.writeValueAsString(s);
    JsonNode jsonNode = OBJECT_MAPPER.readTree(jsonString);

    content.add(jsonNode);

    return s;
}
```

gmsdelmundo: Actually you might be able to use mapper.valueToTree() or mapper.convertValue(object, JsonNode.class)

gmsdelmundo (approved):

PR #159, opened 2023-11-10
differentiate 401 with 500

yishi-ttd:
The error handling logic is on core: https://github.com/IABTechLab/uid2-core/blob/c62c6400a50f121996dd00d85957d59940bebd00/src/main/java/com/uid2/core/vertx/CoreVerticle.java#L237
- If unhandled exception throws, or `AyncResult` failed, it will return 500
- If `AsyncResult` succeeds, but the `AttestationResult` failed, it will return 401 with reason.
- Otherwise, if `AttestationResult` succeeds, it will return 200.

It would be an overkill to define all error enum in `AttestationFailure`. Hence I added addtional `AttestationClientException` field in `AttestationResult`. If it's not null, we will get failure reason from its error message.

This change will impact new GCP attestation provider and new Azure attestation provider. Other attestation provider will not be impacted.

### DIFF
--- a/src/main/java/com/uid2/shared/secure/AttestationResult.java
+++ b/src/main/java/com/uid2/shared/secure/AttestationResult.java
@@ -36,7 +36,7 @@ public class AttestationResult {
     public AttestationFailure getFailure() { return this.failure; }
 
     public String getReason() {
-        if(this.attestationClientException != null){
+        if (this.attestationClientException != null) {
             return this.attestationClientException.getMessage();
         }
         return this.failure.explain();
--- a/src/main/java/com/uid2/shared/secure/AttestationResult.java
+++ b/src/main/java/com/uid2/shared/secure/AttestationResult.java
@@ -6,27 +6,27 @@ public class AttestationResult {
 
     private final String enclaveId;
 
-    private final AttestationClientException attestationException;
+    private final AttestationClientException attestationClientException;
 
     public AttestationResult(AttestationFailure reasonToFail) {
         this.failure = reasonToFail;
         this.publicKey = null;
         this.enclaveId = "Failed attestation, enclave Id unknown";
-        this.attestationException = null;
+        this.attestationClientException = null;
     }
 
     public AttestationResult(AttestationClientException exception) {
         this.failure = AttestationFailure.OTHER;
         this.publicKey = null;
         this.enclaveId = "Failed attestation, enclave Id unknown";
-        this.attestationException = exception;
+        this.attestationClientException = exception;
     }
 
     public AttestationResult(byte[] publicKey, String enclaveId) {
         this.failure = AttestationFailure.NONE;
         this.publicKey = publicKey;
         this.enclaveId = enclaveId;
-        this.attestationException = null;
+        this.attestationClientException = null;
     }
 
     public boolean isSuccess() {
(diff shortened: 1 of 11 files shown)


src/main/java/com/uid2/shared/secure/GcpOidcAttestationProvider.java
```diff
@@ -41,11 +41,14 @@ public void attest(byte[] attestationRequest, byte[] publicKey, Handler<AsyncRes
             if (enclaveId != null) {
                 handler.handle(Future.succeededFuture(new AttestationResult(publicKey, enclaveId)));
             } else {
-                throw new AttestationException("unauthorized token");
+                handler.handle(Future.succeededFuture(new AttestationResult(AttestationFailure.FORBIDDEN_ENCLAVE)));
```
thomasm-ttd, line 44: should probably log this too

src/main/java/com/uid2/shared/secure/azurecc/MaaTokenSignatureValidator.java
```diff
@@ -65,7 +66,7 @@ public MaaTokenPayload validate(String tokenString) throws AttestationException
                 tokenVerifier.verify(tokenString);
             }
         } catch (TokenVerifier.VerificationException e) {
-            throw new AttestationException("Fail to validate the token signature, error: " + e.getMessage());
+            throw new AttestationClientException("Fail to validate the token signature, error: " + e.getMessage());
```
thomasm-ttd, line 69: Does this message get returned to the caller? What does it contain?

thomasm-ttd (commented):
Not finished my review, but some things to look at

src/main/java/com/uid2/shared/secure/AttestationResult.java
```diff
@@ -25,6 +36,9 @@ public boolean isSuccess() {
     public AttestationFailure getFailure() { return this.failure; }
 
     public String getReason() {
+        if(this.attestationClientException != null){
```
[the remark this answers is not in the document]
yishi-ttd: Will fix. Forget to enable "format on save" on my another machine.

src/main/java/com/uid2/shared/secure/azurecc/MaaTokenSignatureValidator.java
yishi-ttd: It will return to caller. No confidential data will be exposed.

Error message could be:
- token expired
- audience not match
- algorithm not supported
- keyid not found
- signature validation failed
- ...

image

src/main/java/com/uid2/shared/secure/GcpOidcAttestationProvider.java
```diff
@@ -41,11 +41,15 @@ public void attest(byte[] attestationRequest, byte[] publicKey, Handler<AsyncRes
             if (enclaveId != null) {
                 handler.handle(Future.succeededFuture(new AttestationResult(publicKey, enclaveId)));
             } else {
-                throw new AttestationException("unauthorized token");
+                LOGGER.warn("Can not find registered gcp-oidc enclave id.");
```
thomasm-ttd, line 44: Can we mirror the logging in the AzureCC one - log the successful enclave id as an info and the unregistered one as a warn, with the enclaveId?

thomasm-ttd (approved):
One small logging comment, otherwise approved

PR #160, opened 2023-12-12
sch-UID2-2470 add roles to service links

sophia-chen-ttd:
Allow Service Links to support roles by adding role parameter to Service Link model. Changed model and parser to use object mapper pattern. If role is null, then an empty hashset is created. Tested creation of service links with empty role, single role, multiple roles and null value.

### DIFF
--- a/src/main/java/com/uid2/shared/model/ServiceLink.java
+++ b/src/main/java/com/uid2/shared/model/ServiceLink.java
@@ -1,8 +1,17 @@
 package com.uid2.shared.model;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
 
 import java.util.Objects;
+import com.uid2.shared.auth.Role;
+
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Set;
 
 public class ServiceLink {
     @JsonProperty("site_id")
@@ -12,12 +21,20 @@ public class ServiceLink {
     @JsonProperty("link_id")
     private String linkId;
     private String name;
+    private Set<Role> roles;
 
-    public ServiceLink(String linkId, int serviceId, int siteId, String name) {
+    @JsonCreator
+    public ServiceLink(
+            @JsonProperty("link_id") String linkId,
+            @JsonProperty("service_id") int serviceId,
+            @JsonProperty("site_id") int siteId,
+            @JsonProperty("name") String name,
+            @JsonProperty("roles") Set<Role> roles) {
         this.linkId = linkId;
         this.serviceId = serviceId;
         this.siteId = siteId;
         this.name = name;
+        this.roles = Objects.requireNonNullElseGet(roles, HashSet::new);
     }
 
     public String getLinkId() {
@@ -39,8 +56,17 @@ public class ServiceLink {
     public String getName() {
         return name;
     }
+
     public void setName(String name) {
         this.name = name;
+    }
+
+    public Set<Role> getRoles() {
+        return roles;
+    }
+
+    public void setRoles(Set<Role> roles) {
+        this.roles = Objects.requireNonNullElseGet(roles, HashSet::new);
     }
 
     @Override
@@ -50,6 +76,7 @@ public class ServiceLink {
                 ", serviceId=" + serviceId +
                 ", linkId='" + linkId + '\'' +
                 ", name='" + name + '\'' +
+                ", roles=" + roles +
                 '}';
     }
 
@@ -58,11 +85,12 @@ public class ServiceLink {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         ServiceLink serviceLink = (ServiceLink) o;
-        return siteId == serviceLink.siteId && serviceId == serviceLink.serviceId && linkId.equals(serviceLink.linkId) && name.equals(serviceLink.name);
+        return siteId == serviceLink.siteId && serviceId == serviceLink.serviceId && linkId.equals(serviceLink.linkId)
+                && name.equals(serviceLink.name) && roles.equals(serviceLink.roles);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(siteId, serviceId, linkId, name);
+        return Objects.hash(siteId, serviceId, linkId, name, roles.hashCode());
     }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/com/uid2/shared/store/parser/ServiceLinkParser.java
```diff
@@ -1,31 +1,28 @@
...
 import java.io.InputStream;
 import java.util.*;
+import java.util.stream.Collectors;
 
 public class ServiceLinkParser implements Parser<Map<String, ServiceLink>> {
 
+    private static final ObjectMapper OBJECT_MAPPER = Mapper.getInstance();
+
+    static {
+        OBJECT_MAPPER.enable(DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE);
```
gmsdelmundo, line 19: Are we using default values for an enum?

src/test/java/com/uid2/shared/store/RotatingServiceLinkStoreTest.java
```diff
@@ -77,20 +80,45 @@ public void loadContentMultipleServices() throws Exception {
...
 
         final long count = serviceLinkStore.loadContent(makeMetadata("locationPath"));
 
-        ServiceLink sl = serviceLinkStore.getServiceLink(1, "abc123");
-        assertNotNull(sl);
-        assertEquals("Test Service 1", sl.getName());
-        assertEquals(1, sl.getServiceId());
-        assertEquals(123, sl.getSiteId());
-        assertEquals("abc123", sl.getLinkId());
+        ServiceLink sl1 = serviceLinkStore.getServiceLink(1, "abc123");
```
gmsdelmundo, line 91: The below assertions can be improved by simplifying the equality checks.
* You currently have the four service links l1, l2, l3, l4. You can put all of these into a list `expected`
* Do `serviceLinkStore.getServiceLink(..., ...)` for each entry, and put them into a list `actual`
* Directly compare the equality of the `expected` and `actual` list. This also ensures your implementation of `ServiceLink.equals()` is correct

src/test/java/com/uid2/shared/store/RotatingServiceLinkStoreTest.java
```diff
@@ -77,20 +80,45 @@ public void loadContentMultipleServices() throws Exception {
...
+
+        ServiceLink sl4 = serviceLinkStore.getServiceLink(3, "jkl1011");
+        assertNotNull(sl4);
+        assertEquals("test2", sl4.getName());
+        assertEquals(3, sl4.getServiceId());
+        assertEquals(124, sl4.getSiteId());
+        assertEquals("jkl1011", sl4.getLinkId());
+        assertEquals(new HashSet<Role>(), sl4.getRoles());
 
         assertNull(serviceLinkStore.getServiceLink(4, "missing"));
```
gmsdelmundo, line 123: This should be its own unit test

src/main/java/com/uid2/shared/store/parser/ServiceLinkParser.java
sophia-chen-ttd: Good point! I just copied the object mapper pattern from clientkey and didn't look into the usage. I've looked at the documentation now, and we don't need it. Thanks for picking it up!

src/main/java/com/uid2/shared/model/ServiceLink.java
```diff
@@ -12,12 +17,20 @@ public class ServiceLink {
     @JsonProperty("link_id")
     private String linkId;
     private String name;
+    private Set<Role> roles;
```
atarassov-ttd, line 20: Does anything need to be done to make this forward-compatible? E.g. what happens if a future json contains a role this binary does not yet know about?

sophia-chen-ttd: I've enabled the 'DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE' for ObjectMapper in ServiceLinkParser again as that allows the json to parse unknown roles as Role.UNKNOWN.

src/main/java/com/uid2/shared/store/parser/ServiceLinkParser.java
sophia-chen-ttd: I realised later that we do need it for roles. There is a default value 'UNKNOWN' already set up in Roles which allows the json creator to use that if the role parsed in is unknown.

atarassov-ttd (approved):

src/test/java/com/uid2/shared/store/RotatingServiceLinkStoreTest.java
```diff
@@ -62,36 +65,46 @@ public void loadContentEmptyArray() throws Exception {
...
-        assertEquals("abc123", sl.getLinkId());
+        assertEquals(Arrays.asList(serviceLinkStore.getServiceLink(1, "abc123"),
+                serviceLinkStore.getServiceLink(2, "abc123"),
+                serviceLinkStore.getServiceLink(1, "ghi789")), Arrays.asList(l1, l2, l3));
 
         assertNull(serviceLinkStore.getServiceLink(4, "missing"));
     }
+
+    @Test
+    public void createService_NullRole_CreatesServiceLinkWithEmptySetOfRoles() throws Exception {
```
gmsdelmundo, line 100: I think the letter after each _ should be lowercase, i.e. createService_nullRole_createsServiceLinkWithEmptySetOfRoles

Same comment with other tests

src/test/java/com/uid2/shared/store/RotatingServiceLinkStoreTest.java
```diff
@@ -62,36 +65,46 @@ public void loadContentEmptyArray() throws Exception {
...
 
         final long count = serviceLinkStore.loadContent(makeMetadata("locationPath"));
 
-        ServiceLink sl = serviceLinkStore.getServiceLink(1, "abc123");
-        assertNotNull(sl);
-        assertEquals("Test Service 1", sl.getName());
-        assertEquals(1, sl.getServiceId());
-        assertEquals(123, sl.getSiteId());
-        assertEquals("abc123", sl.getLinkId());
+        assertEquals(Arrays.asList(serviceLinkStore.getServiceLink(1, "abc123"),
```
gmsdelmundo, line 92: I think just to make it a bit clearer, this can be changed to:
```
List<ServiceLink> expected = List.of(l1, l2, l3);
List<ServiceLink> actual = List.of(
  serviceLinkStore.getServiceLink(1, "abc123"),
  serviceLinkStore.getServiceLink(2, "abc123"),
  serviceLinkStore.getServiceLink(1, "ghi789")
);
assertEquals(expected, actual);
```

gmsdelmundo (approved):

gmsdelmundo (approved):

PR #172, opened 2024-01-17
Fixed issue where null JWT sent to Core on startup

thomasm-ttd:
Added unit test
Tested in unit tests locally
Tested in E2E tests locally
Tested in a branch deploy of both Operator and Optout

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>com.uid2</groupId>
     <artifactId>uid2-shared</artifactId>
-    <version>6.0.14-SNAPSHOT</version>
+    <version>6.0.18-SNAPSHOT</version>
     <name>${project.groupId}:${project.artifactId}</name>
     <description>Library for all the shared uid2 operations</description>
     <url>https://github.com/IABTechLab/uid2docs</url>
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>com.uid2</groupId>
     <artifactId>uid2-shared</artifactId>
-    <version>6.0.10-208e0b52b2</version>
+    <version>6.0.14-SNAPSHOT</version>
     <name>${project.groupId}:${project.artifactId}</name>
     <description>Library for all the shared uid2 operations</description>
     <url>https://github.com/IABTechLab/uid2docs</url>
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>com.uid2</groupId>
     <artifactId>uid2-shared</artifactId>
-    <version>6.0.18-SNAPSHOT</version>
+    <version>6.0.20-SNAPSHOT</version>
     <name>${project.groupId}:${project.artifactId}</name>
     <description>Library for all the shared uid2 operations</description>
     <url>https://github.com/IABTechLab/uid2docs</url>
--- a/src/main/java/com/uid2/shared/attest/UidCoreClient.java
+++ b/src/main/java/com/uid2/shared/attest/UidCoreClient.java
@@ -132,6 +132,8 @@ public class UidCoreClient implements IUidCoreClient, DownloadCloudStorage {
         String jwtToken = this.getJWT();
         if (jwtToken != null && !jwtToken.isBlank()) {
             headers.put(Const.Attestation.AttestationJWTHeader, jwtToken);
+        } else {
+            LOGGER.warn("getJWT returned an empty or null string for the JWT");
         }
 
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>com.uid2</groupId>
     <artifactId>uid2-shared</artifactId>
-    <version>6.0.10-208e0b52b2</version>
+    <version>6.0.14-SNAPSHOT</version>
```
gmsdelmundo, line 9: Don't forget to change this back

src/test/java/com/uid2/shared/attest/UidCoreClientTest.java
```diff
@@ -59,6 +55,37 @@ public void Download_Succeed_RequestSentWithExpectedParameters() throws IOExcept
         when(mockHttpClient.get("https://download", expectedHeaders)).thenReturn(mockHttpResponse);
 
         uidCoreClient.download("https://download");
+        verify(mockAttestationTokenRetriever, times(1)).attest();
+        verify(mockHttpClient, times(1)).get("https://download", expectedHeaders);
+    }
+
+    @Test
+    public void Download_With_Attest_Uses_JWT_Second_Call() throws IOException, CloudStorageException, AttestationTokenRetrieverException {
```
gmsdelmundo, line 63: Can the method name be `DownloadWithAttest_UsesJWTSecondCall()`? Or something along those lines

src/test/java/com/uid2/shared/attest/UidCoreClientTest.java
```diff
@@ -59,6 +55,37 @@ public void Download_Succeed_RequestSentWithExpectedParameters() throws IOExcept
         when(mockHttpClient.get("https://download", expectedHeaders)).thenReturn(mockHttpResponse);
 
         uidCoreClient.download("https://download");
+        verify(mockAttestationTokenRetriever, times(1)).attest();
```
gmsdelmundo, line 58: Can you wrap the `verify()`s under an `assertAll()`? Similar comment for the other test

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -108,6 +106,9 @@ private InputStream getWithAttest(String path, String jwtToken) throws IOExcepti
         }
 
         String attestationToken = attestationTokenRetriever.getAttestationToken();
+        if (jwtToken == null || jwtToken.isEmpty()) {
```
gmsdelmundo, line 109: Would it be better to put this just after `attestationTokenRetriever.attest();`? `jwtToken` would only really be null if we haven't attested yet

i.e.

```
        if (!attestationTokenRetriever.attested()) {
            attestationTokenRetriever.attest();
            jwtToken = this.getJWT();
        }
```

pom.xml
thomasm-ttd: This will get fixed by the build

src/test/java/com/uid2/shared/attest/UidCoreClientTest.java
thomasm-ttd: Fixed

src/main/java/com/uid2/shared/attest/UidCoreClient.java
thomasm-ttd: Moved the code, but still left the check. There are other reasons we might not have attested, such as core being unavailable

thomasm-ttd: Removed the parameter and call getJWT just before we need it

src/main/java/com/uid2/shared/attest/UidCoreClient.java
```diff
@@ -137,8 +128,10 @@ private HttpResponse<String> sendHttpRequest(String path, String attestationToke
         if (attestationToken != null && !attestationToken.isBlank()) {
             headers.put(Const.Attestation.AttestationTokenHeader, attestationToken);
         }
-        if (attestationJWT != null && !attestationJWT.isBlank()) {
-            headers.put(Const.Attestation.AttestationJWTHeader, attestationJWT);
+
+        String jwtToken = this.getJWT();
+        if (jwtToken != null && !jwtToken.isBlank()) {
```
vishalegbert-ttd, line 133: Will the JWT be blank at this point? Should we log if it is?

thomasm-ttd: It shouldn't be, but good to log it.

PR #194, opened 2024-02-20
Shared changes to validate attestation URL for GCP

thomasm-ttd:
Tested locally with a deployed GCP operator.

### DIFF
--- a/src/main/java/com/uid2/shared/secure/gcpoidc/PolicyValidator.java
+++ b/src/main/java/com/uid2/shared/secure/gcpoidc/PolicyValidator.java
@@ -44,11 +44,10 @@ public class PolicyValidator implements IPolicyValidator {
                     ENV_OPT_OUT_ENDPOINT
             )
     );
-    private final String attesationUrl;
+    private final String attestationUrl;
 
-    public PolicyValidator(String attesationUrl) {
-
-        this.attesationUrl = attesationUrl;
+    public PolicyValidator(String attestationUrl) {
+        this.attestationUrl = attestationUrl;
     }
 
     @Override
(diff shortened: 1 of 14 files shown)


src/main/java/com/uid2/shared/secure/gcpoidc/PolicyValidator.java
```diff
@@ -120,13 +131,24 @@ private Environment checkEnvOverrides(TokenPayload payload) throws AttestationEx
...
         }
 
         return env;
     }
 
+    private void checkAttestationUrl(HashMap<String, String> optionalEnvOverrides) throws AttestationException {
+        if (!Strings.isNullOrEmpty(optionalEnvOverrides.get(ENV_CORE_ENDPOINT))) {
+            String givenAttestationUrl = optionalEnvOverrides.get(ENV_CORE_ENDPOINT);
+            if (!UrlEquivalenceValidator.areUrlsEquivalent(givenAttestationUrl, this.attestationUrl, LOGGER)) {
+                throw new AttestationClientException("The given attestation URL is unknown. Given URL: " + givenAttestationUrl);
```
cYKatherine, line 147: We are not using `AttestationFailure` here which is inconsistent with the Nitro version. Should we fix that in Nitro?

thomasm-ttd: The nitro version is a bit different, but I have added the AttestationFailure to the AttestationClientException

src/main/java/com/uid2/shared/secure/gcpoidc/PolicyValidator.java
```diff
@@ -30,14 +32,24 @@ public class PolicyValidator implements IPolicyValidator {
...
             ENV_ENVIRONMENT,
-            ENV_OPERATOR_API_KEY_SECRET_NAME,
-            ENV_CORE_ENDPOINT,
-            ENV_OPT_OUT_ENDPOINT
+            ENV_OPERATOR_API_KEY_SECRET_NAME
     );
 
     private static final Map<Environment, List<String>> OPTIONAL_ENV_OVERRIDES_MAP = ImmutableMap.of(
-            Environment.Integration, ImmutableList.of()
+            Environment.Production, ImmutableList.of(
```
cYKatherine, line 39: Do we want to override for production?

cYKatherine (approved):

src/main/java/com/uid2/shared/secure/gcpoidc/PolicyValidator.java
thomasm-ttd: Yes, we do, so that the url is put into the manifest that is signed and passed for attestation. We can then in Core validate that the Core URL is the one we expect.

PR #201, opened 2024-02-26
Use the returned OptOut URL from the attestation request

thomasm-ttd:
All the unit test passed.
Tested with e2e tests locally


src/test/java/com/uid2/shared/attest/UidCoreClientTest.java
```diff
@@ -83,13 +83,13 @@ public void Download_Attest401_AttestCalledTwice() throws CloudStorageException,
         when(mockHttpClient.get(eq("https://download"), any(HashMap.class))).thenReturn(mockHttpResponse);
 
         uidCoreClient.download("https://download");
-        verify(mockAttestationTokenRetriever, times(2)).attest();
+        verify(mockAttestationResponseHandler, times(2)).attest();
     }
 
     @Test
```
cYKatherine, line 89: Maybe add a `getReturnOptoutUrl()` test

src/test/java/com/uid2/shared/attest/UidOptOutClientTest.java
```diff
@@ -1,35 +1,104 @@
...
+        when(mockHttpResponse.body()).thenReturn(expectedResponseBody);
+        when(mockHttpResponse.statusCode()).thenReturn(200);
+        when(mockHttpClient.get(any(), any())).thenReturn(mockHttpResponse);
+
+        when(mockAttestationResponseHandler.attested()).thenReturn(true);
+        when(mockAttestationResponseHandler.getAttestationToken()).thenReturn("testAttToken");
+        when(mockAttestationResponseHandler.getOptOutUrl()).thenReturn("https://core.example.com");
+
+        InputStream is = this.optOutClient.download(path);
+        verify(mockAttestationResponseHandler, atLeastOnce()).getOptOutUrl();
```
cYKatherine, line 81: why is it `atLeastOnce()` here?

src/main/java/com/uid2/shared/attest/UidOptOutClient.java
```diff
@@ -1,30 +1,59 @@
...
+                URL fullUrl = new URL(baseUrl, path);
+                return super.download(fullUrl.toExternalForm());
+            } catch (MalformedURLException e) {
+                LOGGER.error("Unable to parse OptOut URL", e);
+            } catch (Exception e) {
+                // Specifically not logging the exception as it might contain sensitive URLs
+                LOGGER.error("Unexpected error in UidOptOutClient download");
+            }
+        } else {
+            LOGGER.warn("UidOptOutClient attempting to download but OptOutUrl not available");
```
cYKatherine, line 54: It might also reach this else clause because of `attestationResponseHandler == null`

src/test/java/com/uid2/shared/logback/MaskingLokiJsonEncoderTest.java
```diff
@@ -71,8 +71,8 @@ public void testMaskedExceptionsWithS3(Exception ex, String maskedMessage, Strin
 
         assertAll(
                 "testMaskedExceptionsWithS3",
-                () -> assertEquals(log.split("\n")[0].trim(), maskedMessage),
-                () -> assertTrue(log.split("\n")[1].trim().startsWith(stackTrace))
+                () -> assertEquals(log.split("\n")[0].trim(), maskedMessage, String.format("'%s' must equal '%s'", log.split("\n")[0].trim(), maskedMessage)),
```
cYKatherine, line 74: Where does this change come from?

src/test/java/com/uid2/shared/attest/UidOptOutClientTest.java
thomasm-ttd: It gets called twice, once to check not null, then to get the value.

src/test/java/com/uid2/shared/logback/MaskingLokiJsonEncoderTest.java
thomasm-ttd: This test fails occasionally locally and I can never figure out why as the test just asserts a boolean. By changing the message, we can see why it fails

src/main/java/com/uid2/shared/attest/UidOptOutClient.java
thomasm-ttd: Removed the test for attestationResponseHandler = null as the constructor prevents it being null

src/test/java/com/uid2/shared/attest/UidOptOutClientTest.java
cYKatherine: Should we change `atLeastOnce()` to `times(2)`

PR #214, opened 2024-03-15
UID2-2815 Fix issue where some optout hooks are skipped each time.

lionell-pack-ttd:


lionell-pack-ttd:
Prior to change - there are gaps in the sending.
no changes

After the change - no gaps.
publish change

vishalegbert-ttd (approved):

PR #217, opened 2024-04-08
expose salt snapshot expiration

Ian-Nara:


src/main/java/com/uid2/shared/store/ISaltProvider.java
```diff
@@ -15,6 +15,7 @@ static interface ISaltSnapshot {
         SaltEntry getRotatingSalt(byte[] identity);
 
         String getFirstLevelSalt();
+        Instant getExpires();
```
cody-constine-ttd, line 18: I don't see this added as override in child classes

src/main/java/com/uid2/shared/store/RotatingSaltProvider.java
```diff
@@ -121,7 +121,7 @@ public ISaltSnapshot getSnapshot(Instant asOf) {
             if (!snapshot.isEffective(asOf)) break;
             current = snapshot;
         }
-        return current;
+        return current != null ? current : snapshots.get(snapshots.size()-1);
```
thomasm-ttd, line 124: What if snapshots.size == 0? 
Spacing around the the -1
I would expect to see some test cases for this

Ian-Nara: Added test and subsequently fixed a problem. 

We call loadContent() in operator's Main.java before getSnapshot is ever called so I think it should not be zero, unless our salt file really is empty.

thomasm-ttd (approved):

PR #225, opened 2024-04-16
UID2-2830 add site new app names

caroline-ttd:
uid2-shared: https://github.com/IABTechLab/uid2-shared/pull/225
uid2-admin: https://github.com/IABTechLab/uid2-admin/pull/264

```
{
  "id": 1004,
  "name": "kjhkjhjkm,n,n",
  "description": "",
  "enabled": false,
  "clientTypes": [],
  "domain_names": [
    "mbmnbmnb.com"
  ],
  "app_names": [
    "hgjkhljkjkbh"
  ],
  "visible": true,
  "created": 1713224480,
  "appnNames": [
    "hgjkhljkjkbh"
  ]
}
```


src/main/java/com/uid2/shared/model/Site.java
```diff
@@ -123,6 +157,7 @@ public String toString() {
                 ", description=" + description +
                 ", enabled=" + enabled +
                 ", domain_names=" + domainNames.toString() +
+                ", app_names=" + domainNames.toString() +
```
mcollins-ttd, line 160: Should be `appNames.toString()`

src/main/java/com/uid2/shared/model/Site.java
```diff
@@ -71,6 +76,19 @@ public Site(int id, String name, Boolean enabled, Set<ClientType> types, Set<Str
         this.enabled = enabled;
         this.clientTypes = (types != null) ? new HashSet<>(types) : new HashSet<>();
         this.domainNames = (domains != null) ? new HashSet<>(domains) : new HashSet<>();
+        this.appNames = new HashSet<>();
+        this.visible = DEFAULT_VISIBLE;
+        this.created = Instant.now().getEpochSecond();
+    }
+
+    public Site(int id, String name, Boolean enabled, Set<ClientType> types, Set<String> domains, Set<String> appNames) {
```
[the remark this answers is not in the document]
jon8787: @caroline-ttd could you please look at using constructor chaining?

src/main/java/com/uid2/shared/model/Site.java
```diff
@@ -113,6 +146,7 @@ public Site(int id, String name, String description, Boolean enabled, Set<Client
 
     public void setEnabled(Boolean enabled) { this.enabled = enabled; }
     public void setDomainNames(Set<String> domainNames) { this.domainNames = domainNames; }
+    public void setAppNames(Set<String> appNames) { this.appNames = appNames; }
```
jon8787, line 149: This allows appNames to be null, but the ctor doesn't alllow it. 

Maybe we can make this setter do 
`this.appNames = (appNames != null) ? appNames : new HashSet<>();`
and then the constructor calls this setter instead?

src/main/java/com/uid2/shared/model/Site.java
jon8787: @caroline-ttd we're using Lombok elsewhere in shared so you could consider using that instead

sunnywu: could you look at adding a unit test that could have caught this bug if it's not too complicated?

asloobq: I didn't find a good example of Lombok in our Shared repo, its mainly enums using Lombok at this point. Using Lombok's `@Data` should be enough to replace the getters, setters, ToString and equals method. The `hashCode` method looks different than Lombok's implementation and I am not sure where we are using this

mcollins-ttd: @asloobq I don't think we're actually using these methods outside of test code. I've been asking that we stop adding overrides to new classes unless there's a good reason for them.

src/main/java/com/uid2/shared/model/Site.java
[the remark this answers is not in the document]
caroline-ttd: Yeah, I added the constructor chaining.

src/main/java/com/uid2/shared/model/Site.java
caroline-ttd: I added a new simple test for toString() and use `@Data` to use setters and getters. I keep the `toString()` and `hashCode()` as it is different than Lombok's implementation.

jon8787: Different in what way? Maybe the Lombok implementation is better?
The idea is to prevent bugs due to missing a field.

caroline-ttd: In our code, the` toString()` method presents data in the format `Site{id=1, name='NAME', ...}`, while in Lombok, it's displayed as `Site(id=1, name='NAME', ...)`. These representations differ in two aspects: 1. The original format employs single quotes (') around the name field, which is absent in the Lombok version. 2. The original format utilizes curly braces `{}`, whereas Lombok uses parentheses `()`.

jon8787: Sure, we don't need to retain the existing behavior though. This is just used in logging right (if it's used at all)?

src/main/java/com/uid2/shared/model/Site.java
caroline-ttd: I agree it is good to avoid the duplicate code, but after some further consideration and I got some new ideas. Maybe it is good to keep the format as assigning the value directly in the constructor?

1. Clarity and Intent: Constructors are typically used to set initial values directly, making it clear and explicit what values are being initialized when an object is created. Calling setter methods within a constructor could potentially confuse readers of the code about where initializations are happening.
2. Format: It ensures consistency in the constructor format, streamlining it to one single line of code which is not complicated.

jon8787: It's not complicated to duplicate this code, but it did lead to this bug / inconsistency (allowing null in the setter but not in the constructor). I think it makes sense to DRY if the setter does anything more than `this.x = x`

caroline-ttd: Sure, I will make the change. I feel both way work for me. I do not have a strong preference.

src/main/java/com/uid2/shared/model/Site.java
caroline-ttd: If the `toString() `output format is pre-defined, I will remove this method and let Lombok handles it.

src/main/java/com/uid2/shared/model/Site.java
```diff
@@ -134,11 +99,11 @@ public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         Site site = (Site) o;
-        return id == site.id && name.equals(site.name) && description.equals(site.description) && enabled.equals(site.enabled) && domainNames.equals(site.domainNames) && clientTypes.equals(site.clientTypes) && Objects.equals(visible, site.visible) && created == site.created;
+        return id == site.id && name.equals(site.name) && description.equals(site.description) && enabled.equals(site.enabled) && appNames.equals(site.appNames) && clientTypes.equals(site.clientTypes) && Objects.equals(visible, site.visible) && created == site.created && domainNames.equals(site.domainNames);
```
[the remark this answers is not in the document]
caroline-ttd: To use the Lombok for the equals, I need to remove the hashCode implementation which is different with Lombok. Please kindly let me know if there is any potential risk of changing the `hashcode()` implementation of Site class, I will add it back. @jon8787 @sunnywu @mcollins-ttd @asloobq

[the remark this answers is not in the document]
jon8787: I don't see a risk in changing hashCode() to use Lombok instead. As Matt said, we're probably not even using it. And if we are, we're not dependent on it giving the same hash for different versions of uid2-shared.

src/test/java/com/uid2/shared/store/SiteTest.java
```diff
@@ -1,23 +1,32 @@
...
         assertTrue(a.equals(b));
         assertFalse(a.equals(c));
         assertFalse(c.equals(a));
         assertTrue(c.equals(c));
     }
+
+    @Test
+    public void testToString() {
+        Site a = new Site(1, "1 name", "1 desc", true, new HashSet<>(Collections.singletonList(ClientType.DSP)), new HashSet<>(Collections.singletonList("test.com")), new HashSet<>(Arrays.asList("123456789", "com.123.game.app.android")), true, 0L);
+        String expected = "Site(id=1, name=1 name, description=1 desc, enabled=true, domainNames=[test.com], appNames=[123456789, com.123.game.app.android], clientTypes=[DSP], visible=true, created=0)";
```
sunnywu, line 29: would suggest to make `com.123.game.app.android` case sensitive such as `com.123.Game.app.Android` to confirm the case sensitivity

src/test/java/com/uid2/shared/store/SiteTest.java
```diff
@@ -1,23 +1,32 @@
...
 
 public class SiteTest {
     @Test
     public void testEquals() {
-        Site a = new Site(1, "1 name", "1 desc", true, new HashSet<>(), new HashSet<>(), true, 0L);
-        Site b = new Site(1, "1 name", "1 desc", true, new HashSet<>(), new HashSet<>(), true, 0L);
-        Site c = new Site(1, "1 name", "1 desc", true, new HashSet<>(), new HashSet<>(), null, 0L);
+        Site a = new Site(1, "1 name", "1 desc", true, new HashSet<>(), new HashSet<>(), new HashSet<>(), true, 0L);
+        Site b = new Site(1, "1 name", "1 desc", true, new HashSet<>(), new HashSet<>(), new HashSet<>(), true, 0L);
+        Site c = new Site(1, "1 name", "1 desc", true, new HashSet<>(), new HashSet<>(), new HashSet<>(), null, 0L);
```
sunnywu, line 18: not directly related a result of your work but would be great if you could beef up the equal tests here so we should have `equal` test cases for checking non empty client types/domain name/app name sets too

src/main/java/com/uid2/shared/model/Site.java
```diff
@@ -38,107 +40,40 @@ public Site(@JsonProperty("id") int id,
...
-        return id == site.id && name.equals(site.name) && description.equals(site.description) && enabled.equals(site.enabled) && domainNames.equals(site.domainNames) && clientTypes.equals(site.clientTypes) && Objects.equals(visible, site.visible) && created == site.created;
+    public Site(int id, String name, String description, Boolean enabled, Set<ClientType> types, Set<String> domains, Set<String> appNames, Boolean visible) {
+        this(id, name, description, enabled, types, domains, appNames, visible, Instant.now().getEpochSecond());
     }
 
-    @Override
-    public int hashCode() {
-        return Objects.hash(id, name, description, enabled, domainNames,Arrays.hashCode(clientTypes.toArray()), visible, created);
-    }
+    public Boolean isEnabled() { return enabled; }
```
asloobq, line 76: these getter methods can be generated by lombok, right ?

src/main/java/com/uid2/shared/model/Site.java
asloobq: One way to avoid this inconsistency of ctor vs Setter is to use the `@Nonnull` annotation on the field. Lombok will add an explicit Null Check in the Setter method.

> A number of annotations from popular libraries that indicate non-nullness, such as javax.annotation.Nonnull, if present on the field, result in an explicit null check in the generated setter.

https://projectlombok.org/features/GetterSetter
The responsibility now falls on the caller to not set a null value for the field

asloobq (commented):
may be a good idea to locally package and import into admin and operator to confirm it doesn't break any existing APIs.

src/main/java/com/uid2/shared/model/Site.java
caroline-ttd: Lombok generated the method `getEnabled()` but we use `isEnabled()`

caroline-ttd:
> may be a good idea to locally package and import into admin and operator to confirm it doesn't break any existing APIs.

That is a good idea. I tested locally with admin already. But I have not tested operator yet.

jon8787:
removed 60 lines of code from `Site.java`, great result!

src/test/java/com/uid2/shared/store/SiteTest.java
```diff
@@ -1,23 +1,39 @@
...
+import static org.junit.jupiter.api.Assertions.*;
 
 public class SiteTest {
     @Test
     public void testEquals() {
-        Site a = new Site(1, "1 name", "1 desc", true, new HashSet<>(), new HashSet<>(), true, 0L);
-        Site b = new Site(1, "1 name", "1 desc", true, new HashSet<>(), new HashSet<>(), true, 0L);
-        Site c = new Site(1, "1 name", "1 desc", true, new HashSet<>(), new HashSet<>(), null, 0L);
+        Site a = new Site(1, "1 name", "1 desc", true, new HashSet<>(Collections.singletonList(ClientType.DSP)), new HashSet<>(Collections.singletonList("test.com")), new HashSet<>(Arrays.asList("123456789", "com.123.Game.app.Android")), true, 0L);
+        Site b = new Site(1, "1 name", "1 desc", true, new HashSet<>(Collections.singletonList(ClientType.DSP)), new HashSet<>(Collections.singletonList("test.com")), new HashSet<>(Arrays.asList("123456789", "com.123.Game.app.Android")), true, 0L);
```
sunnywu, line 17: minor: would be good to have a `Site b2` which swap "123456789" and "com.123.Game.app.Android" around and app test that the change of array ordering shouldn't matter for equals method
`assertTrue(a.equals(b2));`

sunnywu (approved):

src/main/java/com/uid2/shared/model/Site.java
asloobq: Oh, ok. Its probably because its a boxed type instead of primitive boolean. In the ctor a default value is always assigned so I don't see a reason for it to keep it boxed. Nitpicking here, I'll leave it upto you :)

asloobq (approved):

src/main/java/com/uid2/shared/model/Site.java
caroline-ttd: I will just leave as it is now :)

caroline-ttd:
Tested locally with operator (`/token/client-generate`, `/key/bidstream` and `/key/sharing`) and admin repo

PR #226, opened 2024-04-16
Ian UI d2 1728 identity buckets oom

Ian-Nara:


thomasm-ttd (approved):

src/main/java/com/uid2/shared/Const.java
```diff
@@ -73,6 +73,8 @@ public static class Config {
         public static final String AwsKmsJwtSigningPublicKeysProp = "aws_kms_jwt_signing_public_keys";
         public static final String EnforceJwtProp = "enforceJwt";
         public static final String MaaServerBaseUrlProp ="maa_server_base_url";
+        public static final String MaxIdentityBucketsResponseEntries = "max_identity_buckets_response_entries";
```
jon8787, line 76: can these go in Operator instead?

PR #281, opened 2024-07-15
add a new function to keep mapping

lizk886:


src/main/java/com/uid2/shared/store/reader/RotatingS3KeyProvider.java
```diff
@@ -10,13 +10,20 @@
 
 import java.time.Instant;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.*;
```
cody-constine-ttd, line 15: Star import is not necessary

cody-constine-ttd (commented):
Small nit otherwise looks good

thomasm-ttd (approved):

PR #290, opened 2024-07-20
improve s3 key manger, add key selection logics

lizk886:
1 improve s3 key manger for unit testing
2 get a more realistic key selection logic

### DIFF
--- a/src/main/java/com/uid2/shared/store/reader/RotatingS3KeyProvider.java
+++ b/src/main/java/com/uid2/shared/store/reader/RotatingS3KeyProvider.java
@@ -97,19 +97,8 @@ public class RotatingS3KeyProvider implements StoreReader<Map<Integer, S3Key>> {
         return siteToKeysMap.getOrDefault(siteId, new ArrayList<>());
     }
 
-    public List<S3Key> getMostRecentKeysForSite(int siteId) {
-        List<S3Key> keys = getKeys(siteId);
-        if (keys.isEmpty()) {
-            throw new IllegalStateException("No keys found for site ID: " + siteId);
-        }
-        return keys.stream()
-                .sorted(Comparator.comparingLong(S3Key::getCreated).reversed())
-                // Get most recently created 3 keys
-                .limit(3)
-                .collect(Collectors.toList());
-    }
-
      public S3Key getEncryptionKeyForSite(Integer siteId) {
+        //get the youngest activated key
         Collection<S3Key> keys = getKeysForSite(siteId);
          long now = Instant.now().getEpochSecond();
          if (keys.isEmpty()) {
--- a/src/main/java/com/uid2/shared/store/reader/RotatingS3KeyProvider.java
+++ b/src/main/java/com/uid2/shared/store/reader/RotatingS3KeyProvider.java
@@ -17,6 +17,7 @@ import java.util.Collection;
 import java.util.Comparator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import java.time.Instant;
 
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


src/main/java/com/uid2/shared/store/reader/RotatingS3KeyProvider.java
```diff
@@ -6,13 +6,9 @@
...
 import io.vertx.core.json.JsonObject;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.ArrayList;
-import java.util.Set;
-import java.util.HashSet;
+
+import java.util.*;
```
cody-constine-ttd, line 10: Do we need star import? Lets avoid this.

src/main/java/com/uid2/shared/store/reader/RotatingS3KeyProvider.java
```diff
@@ -84,21 +82,22 @@ public Collection<S3Key> getKeysForSite(Integer siteId) {
...
-                        largestKey = key;
-                    }
-                }
-            }
-            return largestKey;
         }
+        return findLargestKey(keys);
     }
+
+    S3Key findLargestKey(Collection<S3Key> keys) {
```
cody-constine-ttd, line 97: Do we use this anywhere besides the above function? Seems like we can just get rid of it

src/main/java/com/uid2/shared/store/reader/RotatingS3KeyProvider.java
```diff
@@ -84,21 +82,22 @@ public Collection<S3Key> getKeysForSite(Integer siteId) {
                 .collect(Collectors.toList());
     }
 
+    public List<S3Key> getKeysForSiteFromMap(int siteId) {
```
cody-constine-ttd, line 85: I thinks getKeys is a better name

src/main/java/com/uid2/shared/store/reader/RotatingS3KeyProvider.java
lizk886: `findLargestKey` is currently being used by `getEncryptionKeyForSite`. Right now, I am using a temporary key selection method by simply choosing the largest key for a specific siteId. I planned to use this temporary key selection logic until someone started work on key rotation.

src/main/java/com/uid2/shared/store/reader/RotatingS3KeyProvider.java
```diff
@@ -23,7 +28,7 @@ public class RotatingS3KeyProvider implements StoreReader<Map<Integer, S3Key>> {
     ScopedStoreReader<Map<Integer, S3Key>> reader;
 
     private static final Logger LOGGER = LoggerFactory.getLogger(RotatingS3KeyProvider.class);
-    public final Map<Integer, List<S3Key>> siteToKeysMap = new HashMap<>();
+    Map<Integer, List<S3Key>> siteToKeysMap = new ConcurrentHashMap<>();
```
cody-constine-ttd, line 31: Why is this changed to a concurrentHashMap?

src/main/java/com/uid2/shared/store/reader/RotatingS3KeyProvider.java
```diff
@@ -84,21 +93,31 @@ public Collection<S3Key> getKeysForSite(Integer siteId) {
                 .collect(Collectors.toList());
     }
 
-    public S3Key getEncryptionKeyForSite(Integer siteId) {
-        Collection<S3Key> keys = getKeysForSite(siteId);
+    public List<S3Key> getKeys(int siteId) {
+        return siteToKeysMap.getOrDefault(siteId, new ArrayList<>());
+    }
+
+    public List<S3Key> getMostRecentKeysForSite(int siteId) {
```
cody-constine-ttd, line 100: getActiveKeys

src/main/java/com/uid2/shared/store/reader/RotatingS3KeyProvider.java
```diff
@@ -84,21 +93,31 @@ public Collection<S3Key> getKeysForSite(Integer siteId) {
...
+            throw new IllegalStateException("No keys found for site ID: " + siteId);
+        }
+        return keys.stream()
+                .sorted(Comparator.comparingLong(S3Key::getCreated).reversed())
+                // Get most recently created 3 keys
+                .limit(3)
+                .collect(Collectors.toList());
+    }
+
+     public S3Key getEncryptionKeyForSite(Integer siteId) {
```
cody-constine-ttd, line 112: I think we can combine this with the function above, and always send all active keys.

src/main/java/com/uid2/shared/store/reader/RotatingS3KeyProvider.java
lizk886: I am a little bit worried that there would be a concurrency issue,  such as simultaneous updates to the siteId map keys and reading keys occurring at the same time. Can revert this back to hashMap if thats not a problem

cody-constine-ttd (approved):

PR #308, opened 2024-08-03
Wzh uid2 3575 encrypted scoped store reader change without knowing site

lizk886:


src/main/java/com/uid2/shared/store/EncryptedScopedStoreReader.java
```diff
@@ -23,12 +23,14 @@
 public class EncryptedScopedStoreReader<T> extends ScopedStoreReader<T> {
     private static final Logger LOGGER = LoggerFactory.getLogger(EncryptedScopedStoreReader.class);
 
-    private final int siteId;
+    private int siteId = -1;
     private final RotatingS3KeyProvider s3KeyProvider;
 
-    public EncryptedScopedStoreReader(DownloadCloudStorage fileStreamProvider, EncryptedScope scope, Parser<T> parser, String dataTypeName, RotatingS3KeyProvider s3KeyProvider) {
```
lizk886, line 29: It is a little bit tricky here, we can benefit from using EncryptedScope so that siteId is passed in and that can be used to generated some new keys for that siteId if there are not decryption keys avaible.

PR #327, opened 2024-10-14
Consolidate the Attestation exceptions to retryable or AttestationFailure

thomasm-ttd:


pom.xml
```diff
@@ -285,6 +285,9 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>3.2.5</version>
+                <configuration>
+                    <argLine>-XX:+EnableDynamicAgentLoading</argLine>
```
Ian-Nara, line 289: What is this for?

src/test/java/com/uid2/shared/attest/UidCoreClientTest.java
```diff
@@ -62,18 +62,18 @@ public void Download_Succeed_RequestSentWithExpectedParameters() throws IOExcept
...
             uidCoreClient.download("https://download");
         });
-        String expectedExceptionMessage = "download https://download error: http status: 401, test failure";
+        String expectedExceptionMessage = "download error: AttestationResponseCode: AttestationFailure, test failure";
         assertEquals(expectedExceptionMessage, result.getMessage());
     }
 
     @Test
-    public void Download_Attest401_AttestCalledTwice() throws CloudStorageException, IOException, InterruptedException, AttestationResponseHandlerException {
+    public void Download_Attest401_getOptOut_NotCalled() throws CloudStorageException, IOException, AttestationResponseHandlerException {
```
Ian-Nara, line 76: What change is this test update related to?

pom.xml
thomasm-ttd: It removes the warning when the tests start that come from mockito.

src/test/java/com/uid2/shared/attest/UidCoreClientTest.java
thomasm-ttd: I removed the code that calls attest a second time after the first one returns a 401 as that code is not used in production (we log if it is called), and if it returns a 401, calling a second time will get the same response. That test then failed as it checks for 2 calls. Renamed the test as the thing it should be checking is that we don't call getOptOut after getting a 401

Ian-Nara (approved):

PR #351, opened 2024-11-26
Xuy UI d2 2578 add site name to metrics

clarkxuyang:
Description:

Add site name to metrics

Test

Unit test

### DIFF
--- a/src/main/java/com/uid2/shared/vertx/RequestCapturingHandler.java
+++ b/src/main/java/com/uid2/shared/vertx/RequestCapturingHandler.java
@@ -218,16 +218,14 @@ public class RequestCapturingHandler implements Handler<RoutingContext> {
         return null;
     }
 
-    private void incrementMetricCounter(String apiContact, Integer siteId, String siteName,
-                                        String host, int status, HttpMethod method, String path) {
+    private void incrementMetricCounter(String apiContact, Integer siteId, String siteName, String host, int status, HttpMethod method, String path) {
         assert apiContact != null;
         String key = apiContact + "|" + siteId + "|" + siteName + "|" + host + "|" + status + "|" + method.name() + "|" + path;
         if (!_apiMetricCounters.containsKey(key)) {
             Counter counter = Counter
                     .builder("uid2.http_requests")
                     .description("counter for how many http requests are processed per each api contact and status code")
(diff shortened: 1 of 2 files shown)


cYKatherine (approved):

src/main/java/com/uid2/shared/vertx/RequestCapturingHandler.java
```diff
@@ -32,11 +34,21 @@ public class RequestCapturingHandler implements Handler<RoutingContext> {
     private Queue<String> _capturedRequests = null;
     private final Map<String, Counter> _apiMetricCounters = new HashMap<>();
     private final Map<String, Counter> _clientAppVersionCounters = new HashMap<>();
+    private ISiteStore siteStore;
 
     private static String formatRFC1123DateTime(long time) {
         return DateTimeFormatter.RFC_1123_DATE_TIME.format(Instant.ofEpochMilli(time).atZone(ZONE_GMT));
     }
 
+    public RequestCapturingHandler()
```
atarassov-ttd, line 43: Is this constructor still needed?

cYKatherine: I checked it is needed
> Java does not provide a default no-argument constructor if you define any other constructor.

clarkxuyang: It is just a super cautious step for backward compatible - in case someone with old operator wants to use the new shared lib.

atarassov-ttd: At some point after all uses are replace we should get rid of this. Should it be marked as deprecated?

src/test/java/com/uid2/shared/vertx/RequestCapturingHandlerTest.java
```diff
@@ -160,11 +178,11 @@ public void getSiteIdFromRoutingContextData(String key, Object value, String sit
     private static Stream<Arguments> siteIdRoutingContextData() {
         // Arguments are: routing context data key, routing context data value, site ID tag.
```
thomasm-ttd, line 179: Comment needs to be updated

PR #388, opened 2025-02-24
xuy-UID2-4848-improve-clouddownload-logging #387

clarkxuyang:


src/main/java/com/uid2/shared/cloud/URLStorageWithMetadata.java
```diff
@@ -35,15 +36,25 @@ public void upload(InputStream input, String cloudPath) throws CloudStorageExcep
...
+            } else {
+                httpConn = (HttpURLConnection) url.openConnection();
+            }
+            
+            int responseCode = httpConn.getResponseCode();
+            if (responseCode >= 200 && responseCode < 300) {
+                return httpConn.getInputStream();
             } else {
-                return url.openStream();
+                throw new CloudStorageException("Url download error: HTTP response code " + responseCode
```
cody-constine-ttd, line 51: Throwing here does nothing as it will just be caught by the catch below and throw there again.

src/main/java/com/uid2/shared/cloud/URLStorageWithMetadata.java
```diff
@@ -35,15 +36,25 @@ public void upload(InputStream input, String cloudPath) throws CloudStorageExcep
...
-                return url.openStream();
+                throw new CloudStorageException("Url download error: HTTP response code " + responseCode 
+                + " Visit UID2 docs for more details");
             }
         } catch (Throwable t) {
-            // Do not log the message or the original exception as that may contain the pre-signed url
-            throw new CloudStorageException("url download error: " + t.getClass().getSimpleName());
+            // Do not log the original exception as it may contain sensitive information such as the pre-signed URL
+            throw new CloudStorageException("Url download error: " + t.getClass().getSimpleName() +
+                " Visit UID2 docs for more details");
```
abuabraham-ttd, line 57: Can we follow same style as other customer config errors;

EXX: CloudStorageDownloadError: Cannot download required files for startup, <error code>. Please see  https://unifiedid.com/docs/guides/operator-guide-aws-marketplace#uid2-operator-error-codes for more .

src/main/java/com/uid2/shared/cloud/URLStorageWithMetadata.java
clarkxuyang: update, the purpose is to log the response code.

src/main/java/com/uid2/shared/cloud/URLStorageWithMetadata.java
clarkxuyang: Do we want to continue the error code defined in operator, E9? It might be hard to trace.

PR #390, opened 2025-02-27
Allow JWT validation failures when not enforcing JWTs

vishalegbert-ttd:


src/main/java/com/uid2/shared/middleware/AttestationMiddleware.java
```diff
@@ -85,32 +86,36 @@ public void handle(RoutingContext rc) {
...
                 }
             }
 
+            if (!isJwtValid && this.enforceJwt) {
+                success = false;
+            } else if (!isJwtValid && !this.enforceJwt) {
+                LOGGER.info("JWT validation has failed, but JWTs are not being enforced.");
+            }
+
             if (success) {
```
abuabraham-ttd, line 119: why not just add `!isJwtValid && this.enforceJwt` here and not use `success`? 

If !isJwtValid and  enforceJwt: fail
else: log and onFailedAttestation?

asloobq (approved):

src/main/java/com/uid2/shared/middleware/AttestationMiddleware.java
asloobq: we can't ignore the result of `success = attestor.validateToken(userToken, encryptedToken);` above, probably thats why

abuabraham-ttd: oh, I didn't expand. Thanks

PR #393, opened 2025-02-27
Handle incorrect address in X-Forward-Host headers

asloobq:
When the value of `X-Forward-Host` is invalid, Vertx throws a NPE when trying to get the remote address from the request context. The NPE results in a Http Status code 500. To avoid this and return an appropriate 4xx error, handling this NPE gracefully and continuing request processing.

Exceptions seen in logs as 
```
2025-02-25 13:46:55.859	
	at com.uid2.shared.vertx.RequestCapturingHandler.handle(RequestCapturingHandler.java:31)
	2025-02-25 13:46:55.859	
	at com.uid2.shared.vertx.RequestCapturingHandler.handle(RequestCapturingHandler.java:65)
	2025-02-25 13:46:55.859	
	at io.vertx.ext.web.impl.HttpServerRequestWrapper.remoteAddress(HttpServerRequestWrapper.java:154)
	2025-02-25 13:46:55.859	
	at io.vertx.ext.web.impl.ForwardedParser.remoteAddress(ForwardedParser.java:102)
	2025-02-25 13:46:55.859	
	at io.vertx.ext.web.impl.ForwardedParser.calculate(ForwardedParser.java:122)
	2025-02-25 13:46:55.859	
	at io.vertx.ext.web.impl.ForwardedParser.calculateXForward(ForwardedParser.java:184)
	2025-02-25 13:46:55.859	
	at io.vertx.ext.web.impl.ForwardedParser.setHostAndPort(ForwardedParser.java:199)
	2025-02-25 13:46:55.859	
21:46:55.856 [vert.x-eventloop-thread-17] ERROR c.u.o.vertx.GenericFailureHandler - URL: [/static/js/uid2-esp-0.0.1a.js] - Error response code: [500] - Error: java.lang.NullPointerException: Cannot invoke "io.vertx.core.net.HostAndPort.host()" because "authority" is null
```

### DIFF
--- a/src/main/java/com/uid2/shared/vertx/RequestCapturingHandler.java
+++ b/src/main/java/com/uid2/shared/vertx/RequestCapturingHandler.java
@@ -62,11 +62,19 @@ public class RequestCapturingHandler implements Handler<RoutingContext> {
         }
 
         long timestamp = System.currentTimeMillis();
-        String remoteClient = getClientAddress(context.request().remoteAddress());
+        String remoteClient = null;
+        try {
+            SocketAddress remoteAddress = context.request().remoteAddress();
+            remoteClient = getClientAddress(remoteAddress);
+        } catch (NullPointerException ex) {
+            LOGGER.warn("remoteAddress() throws NullPointerException");
+        }
+
         HttpMethod method = context.request().method();
         String uri = context.request().uri();
         HttpVersion version = context.request().version();
-        context.addBodyEndHandler(v -> captureNoThrow(context, timestamp, remoteClient, version, method, uri));
+        String finalRemoteClient = remoteClient;
+        context.addBodyEndHandler(v -> captureNoThrow(context, timestamp, finalRemoteClient, version, method, uri));
         context.next();
     }
 
@@ -80,8 +88,8 @@ public class RequestCapturingHandler implements Handler<RoutingContext> {
     private void captureNoThrow(RoutingContext context, long timestamp, String remoteClient, HttpVersion version, HttpMethod method, String uri) {
         try {
             capture(context, timestamp, remoteClient, version, method, uri);
-        } catch (Throwable t) {
-            LOGGER.error("capture() throws", t);
+        } catch (RuntimeException e) {
+            LOGGER.error("capture() throws", e);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/uid2/shared/vertx/RequestCapturingHandler.java
```diff
@@ -62,11 +62,19 @@ public void handle(RoutingContext context) {
         }
 
         long timestamp = System.currentTimeMillis();
-        String remoteClient = getClientAddress(context.request().remoteAddress());
+        String remoteClient = null;
+        try {
+            SocketAddress remoteAddress = context.request().remoteAddress();
+            remoteClient = getClientAddress(remoteAddress);
+        } catch (NullPointerException ex) {
```
clarkxuyang, line 69: Catching a NullPointerException (NPE) directly is generally not ideal in Java. Instead, it's better to prevent the NPE by checking for null conditions before access

src/main/java/com/uid2/shared/vertx/RequestCapturingHandler.java
```diff
@@ -62,11 +62,19 @@ public void handle(RoutingContext context) {
...
+            LOGGER.warn("remoteAddress() throws NullPointerException");
+        }
+
         HttpMethod method = context.request().method();
         String uri = context.request().uri();
         HttpVersion version = context.request().version();
-        context.addBodyEndHandler(v -> captureNoThrow(context, timestamp, remoteClient, version, method, uri));
+        String finalRemoteClient = remoteClient;
```
clarkxuyang, line 76: what's this additional assignment used for? can we still use remoteClient?

asloobq: The variable used in lambda expression below and it should be final or effectively final

src/main/java/com/uid2/shared/vertx/RequestCapturingHandler.java
asloobq: The problem isn't that the return value is null. The NPE is thrown from within the vertx library.

clarkxuyang: Ignore above, It is actually verx throw the NPE while passing the x-forward header and we have no control over it.

clarkxuyang (approved):

PR #403, opened 2025-03-18
Added verbose logging for S3 download

gmsdelmundo:


src/main/java/com/uid2/shared/cloud/CloudStorageS3.java
```diff
@@ -25,9 +25,10 @@ public class CloudStorageS3 implements TaggableCloudStorage {
 
     private final AmazonS3 s3;
     private final String bucket;
+    private final boolean verbose;
     private long preSignedUrlExpiryInSeconds = 3600;
 
-    public CloudStorageS3(String accessKeyId, String secretAccessKey, String region, String bucket, String s3Endpoint) {
+    public CloudStorageS3(String accessKeyId, String secretAccessKey, String region, String bucket, String s3Endpoint, boolean verbose) {
```
vishalegbert-ttd, line 31: Will we need to update any consuming applications to fix this breaking change? Does this need to be a breaking change?

src/main/java/com/uid2/shared/cloud/CloudStorageS3.java
```diff
@@ -132,7 +137,7 @@ public InputStream download(String cloudPath) throws CloudStorageException {
             return obj.getObjectContent();
         } catch (AmazonS3Exception e) {
             if (e.getErrorCode().equals("NoSuchKey")) {
-                throw new CloudStorageException("The specified key does not exist: " + e.getClass().getSimpleName() + ": " + bucket);
+                throw new CloudStorageException("The specified key does not exist: " + e.getClass().getSimpleName() + ": " + bucket + (verbose ? " - " + e.getMessage() : ""));
             } else {
                 throw new CloudStorageException("s3 get error: " + e.getClass().getSimpleName() + ": " + bucket);
```
vishalegbert-ttd, line 142: Didn't the exception happen here?

src/main/java/com/uid2/shared/cloud/CloudStorageS3.java
gmsdelmundo: No, I added an overloaded constructor below with that retains original behaviour:
```
    public CloudStorageS3(String accessKeyId, String secretAccessKey, String region, String bucket, String s3Endpoint) {
        this(accessKeyId, secretAccessKey, region, bucket, s3Endpoint, false);
    }
```

src/main/java/com/uid2/shared/cloud/CloudStorageS3.java
gmsdelmundo: Originally had the verbose logging in more lines but kept the wrong one

Fixed

src/main/java/com/uid2/shared/cloud/CloudStorageS3.java
gmsdelmundo: It keeps the same signature:
```
public CloudStorageS3(String accessKeyId, String secretAccessKey, String region, String bucket, String s3Endpoint) {
```

vishalegbert-ttd (approved):

PR #424, opened 2025-04-25
Adding new Identity Map V3 and Refreshable UID fields to SaltEntry

aulme:
1. Allow new fields for Identity Map V3 and Refreshable UIDs in salt files
2. Extracted SaltFileParser into a separate class, reduced duplication between encrypted/unencrypted flow, added tests
3. General refactoring


src/main/java/com/uid2/shared/model/SaltEntry.java
```diff
@@ -1,31 +1,43 @@
...
-    public long getLastUpdated() {
-        return lastUpdated;
+        KeyMaterial currentKey,
+        KeyMaterial previousKey
+) {
+    @Override
+    public String toString() {
+        return "SaltEntry{" +
+                "id=" + id +
+                ", hashedId='" + hashedId + '\'' +
```
gmsdelmundo, line 19: `'\''` can just be `"'"`, but why does `hashedId` have to be surrounded by single quotes?

src/main/java/com/uid2/shared/model/SaltEntry.java
```diff
@@ -1,31 +1,43 @@
...
+                '}';
     }
 
-    public String getSalt() {
-        return salt;
+    public record KeyMaterial(
+            int id,
+            String key,
+            String salt
+    ){
```
gmsdelmundo, line 33: Space between `)` and `{` for consistency with the above

src/main/java/com/uid2/shared/model/SaltEntry.java
```diff
@@ -1,31 +1,43 @@
...
-public class SaltEntry {
-    private final long id;
-    private final String hashedId;
-    private final long lastUpdated;
-    private final String salt;
+public record SaltEntry(
+        long id,
+        String hashedId,
+        long lastUpdated,
+        String salt,
```
gmsdelmundo, line 7: Do we want to rename this to `currentSalt` for consistency with `currentKey`?

src/main/java/com/uid2/shared/model/SaltEntry.java
```diff
@@ -1,31 +1,43 @@
...
+        String previousSalt,
 
-    public long getLastUpdated() {
-        return lastUpdated;
+        KeyMaterial currentKey,
+        KeyMaterial previousKey
+) {
+    @Override
+    public String toString() {
+        return "SaltEntry{" +
```
gmsdelmundo, line 17: Is the reason for this representation of `toString()` so that we can have a JSON-formatted string that Loki can pick up? If so, don't all the values need to be surrounded with double quotes?

We already represent some things in `logfmt` format (mostly explicitly done when we log stuff out), should we make it consistent?

src/main/java/com/uid2/shared/store/salt/SaltFileParser.java
```diff
@@ -0,0 +1,60 @@
...
+            if (fields.length > 7) {
+                currentKey = new SaltEntry.KeyMaterial(Integer.parseInt(fields[5]), fields[6], fields[7]);
+            }
+            if (fields.length > 10) {
+                previousKey = new SaltEntry.KeyMaterial(Integer.parseInt(fields[8]), fields[9], fields[10]);
+            }
+
+            return new SaltEntry(id, hashedId, lastUpdated, salt, refreshFrom, previousSalt, currentKey, previousKey);
+        } catch (Exception e) {
+            throw new RuntimeException("Trouble parsing Salt Entry " + line, e);
```
gmsdelmundo, line 56: If for some reason this throws an error, but the salt fields somehow parsed successfully, won't the salts get logged as plaintext with this logging?

gmsdelmundo: Also I like how you used "Trouble" here 😆

src/main/java/com/uid2/shared/store/salt/SaltFileParser.java
```diff
@@ -0,0 +1,60 @@
...
+            final String hashedId = this.idHashingScheme.encode(id);
+            final long lastUpdated = Long.parseLong(fields[1]);
+            final String salt = fields[2];
+
+            Long refreshFrom = null;
+            String previousSalt = null;
+            SaltEntry.KeyMaterial currentKey = null;
+            SaltEntry.KeyMaterial previousKey = null;
+
+            if (fields.length > 3) {
```
gmsdelmundo, line 41: Do we want to make a note here that all these if-statements can be removed once everything has been rolled out? Like a `TODO` comment

src/main/java/com/uid2/shared/store/salt/SaltFileParser.java
```diff
@@ -0,0 +1,60 @@
...
+import com.uid2.shared.model.SaltEntry;
+
+public class SaltFileParser {
+    private final IdHashingScheme idHashingScheme;
+
+    public SaltFileParser(IdHashingScheme idHashingScheme) {
+        this.idHashingScheme = idHashingScheme;
+    }
+
+    public SaltEntry[] parseFile(String saltFileContent, Integer size) {
```
gmsdelmundo, line 12: Does this get used anywhere except from tests?

src/main/java/com/uid2/shared/model/SaltEntry.java
aulme: That's the default toString generated by IntelliJ. It's not actually multiline if you look at it - the lines are just to break it down. I think it has some sort of toString format in mind, id is an int so doesn't get quotes, hashedId is string so it does.

src/main/java/com/uid2/shared/model/SaltEntry.java
aulme: Sure

src/main/java/com/uid2/shared/model/SaltEntry.java
aulme: sure

src/main/java/com/uid2/shared/model/SaltEntry.java
aulme: Again, this is InteiiJ's convention on how to do toString https://www.jetbrains.com/help/idea/generating-code.html#generate-tostring

src/main/java/com/uid2/shared/store/salt/SaltFileParser.java
aulme: That's exactly why I've added toString override and <REDACTED> on all sensitive fields of the record.
I think before that change it would have (this is existing code, just extracted)

src/main/java/com/uid2/shared/store/salt/SaltFileParser.java
aulme: Yes, in EncryptedRotatingSaltProvider. The encrypted store uses parseFile, the unencrypted - parseFileLines :/

src/main/java/com/uid2/shared/store/salt/SaltFileParser.java
gmsdelmundo: I guess there hasn't really been an issue with salt file parsing so it hasn't come up

But we're not logging out the `SaltEntry` which would have your `toString()` overrides - we're logging out the raw line itself from the file

src/main/java/com/uid2/shared/model/SaltEntry.java
gmsdelmundo: Understood

src/main/java/com/uid2/shared/store/salt/SaltFileParser.java
aulme: Ah good point. Let's discuss at standup.

Yeah, I've added overrides in SaltEntry specifically so it doesn't randomly show up in logs in general, not just this one. I think a good practice for all of our data types with secret fields.

aulme: I've changed this to log the line number instead of the salt line.

gmsdelmundo (approved):

PR #431, opened 2025-05-01
check if the pod is terminating in the healthcheck

Ian-Nara:


src/main/java/com/uid2/shared/health/HealthManager.java
```diff
@@ -25,7 +33,19 @@ public synchronized HealthComponent registerComponent(String name, boolean initi
     public boolean isHealthy() {
         // simple composite logic: service is healthy if none child component is unhealthy
         List<IHealthComponent> list = this.componentList.get();
-        return list.stream().filter(c -> !c.isHealthy()).count() == 0;
+        boolean componentsHealthy = list.stream().filter(c -> !c.isHealthy()).count() == 0;;
+        return componentsHealthy && !checkPodTerminating();
+    }
+
+    private boolean checkPodTerminating() {
```
mcollins-ttd, line 40: I don't think checking for a pod termination file should be a responsibility of HealthManager. This could be done as its own health check component.

Ian-Nara: If we reuse the pattern of the named components, then that would have us tell health manager by some string ID weather the component is healthy or not at a given time. Then we need some background thread monitoring continuously to update the health component right?

mcollins-ttd: You don't need a background thread, you could use a period timer. A simple implementation:

```java
private void periodicallyCheckForShutdownFile(Duration interval, String file) {
    this.vertx.setPeriodic(interval.toMillis(), l -> {
        if (vertx.fileSystem().existsBlocking(file)) {
            healthComponent.setHealthStatus(false, "shutting down");
        }
    });
}
```

asloobq (approved):

PR #443, opened 2025-05-22
Record component health metrics

aulme:


src/main/java/com/uid2/shared/health/HealthManager.java
```diff
@@ -1,14 +1,27 @@
...
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
 public class HealthManager {
     public static HealthManager instance = new HealthManager();
-    private AtomicReference<List<IHealthComponent>> componentList = new AtomicReference(new ArrayList<IHealthComponent>());
+    private final AtomicReference<List<IHealthComponent>> components = new AtomicReference<>(new ArrayList<>());
+    private MeterRegistry registry = null;
+
+    public synchronized void setMeterRegistry(MeterRegistry newRegistry) {
```
vishalegbert-ttd, line 16: I think we can just use Metrics.globalRegistry?

vishalegbert-ttd (approved):

src/main/java/com/uid2/shared/health/HealthManager.java
aulme: Changed to global.

PR #449, opened 2025-05-27
sch-UID2-5484 Added API instance of object mapper

sophia-chen-ttd:


src/main/java/com/uid2/shared/util/Mapper.java
```diff
@@ -12,10 +12,15 @@ public final class Mapper {
             .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
             .build();
 
+    private static final ObjectMapper API_INSTANCE = JsonMapper.builder()
```
gmsdelmundo, line 15: This is enabled by default, so you can just do

`private static final ObjectMapper API_INSTANCE = JsonMapper.builder().build();`

gmsdelmundo: Also whitespace between this and `private Mapper() {}`

gmsdelmundo (approved):

PR #457, opened 2025-06-04
Vse UI d2 5593 add service instance id shared

vishalegbert-ttd:


src/main/java/com/uid2/shared/Const.java
```diff
@@ -30,6 +30,8 @@ public static class Config {
         // this file stores integration config values, e.g. it communicates with other uid2 services that runs locally.
         public static final String INTEG_CONFIG_PATH = "conf" + File.separator + "integ-config.json";
 
+        public static final String InstanceIdPrefixProp = "instance_id_prefix";
+
```
abuabraham-ttd, line 34: Consider renaming instance_id_prefix to uid_instance_id_prefix to avoid confusion with the AWS instanceId (e.g., from EC2 metadata). 

In environments like private operators, IMO it’s easy to misinterpret the meaning otherwise.

abuabraham-ttd (approved):

PR #468, opened 2025-06-09
use logfmt for audit logs

vishalegbert-ttd:


src/main/java/com/uid2/shared/audit/Audit.java
```diff
@@ -1,5 +1,6 @@
 package com.uid2.shared.audit;
 
+import com.uid2.shared.util.LogFmtLineBuilder;
```
abuabraham-ttd, line 3: Let's hold off on this. If we can get parsing and labeling through grafana agent, that can work across services, Java/nodeJS, and SAAS solutions we use.

PR #469, opened 2025-06-09
Handle "unknown" in string and use traceID there

abuabraham-ttd:


src/main/java/com/uid2/shared/audit/Audit.java
```diff
@@ -234,6 +234,10 @@ private String defaultIfNull(String s) {
         return s != null ? s : "unknown";
     }
 
+    private String defaultIfNull(String s, String defaultValue) {
```
asloobq, line 237: This is same as Objects.toString, probably better to use that
https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#toString-java.lang.Object-java.lang.String-

asloobq (commented):
please update unit tests as well

vishalegbert-ttd (approved):

PR #475, opened 2025-06-16
add link id regex field to services

Ian-Nara:


src/main/java/com/uid2/shared/model/Service.java
```diff
@@ -65,11 +81,16 @@ public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         Service service = (Service) o;
-        return serviceId == service.serviceId && siteId == service.siteId && name.equals(service.name) && roles.equals(service.roles);
+        boolean compare = serviceId == service.serviceId && siteId == service.siteId && name.equals(service.name) && roles.equals(service.roles);
+        if (linkIdRegex == null || service.linkIdRegex == null) {
+            return compare && linkIdRegex == service.linkIdRegex;
```
mcollins-ttd, line 86: We can use `&& Objects.equals(linkIdRegex, service.linkIdRegex)` to skip the null tests.

src/main/java/com/uid2/shared/model/Service.java
```diff
@@ -30,6 +37,14 @@ public int getSiteId() {
         return siteId;
     }
 
+    public String getLinkIdRegex() {
```
mcollins-ttd, line 40: nit: Can we move the new methods so that `getSiteId` and `setSiteId` are next to each other?

src/main/java/com/uid2/shared/model/Service.java
```diff
@@ -57,6 +72,7 @@ public String toString() {
                 ", siteId=" + siteId +
                 ", name='" + name + '\'' +
                 ", roles=" + roles +
+                ", linkIdRegex=" + linkIdRegex +
```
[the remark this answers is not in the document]
Ian-Nara: no, java handles the concatenation by automatically converting null to "null"

mcollins-ttd (approved):

PR #494, opened 2025-06-26
add disabled field to services and service links

Ian-Nara:


src/main/java/com/uid2/shared/model/Service.java
```diff
@@ -16,12 +16,17 @@ public class Service {
...
     public Service(int serviceId, int siteId, String name, Set<Role> roles) {
-        this(serviceId, siteId, name, roles, null);
+        this(serviceId, siteId, name, roles, null, false);
     }
 
     public Service(int serviceId, int siteId, String name, Set<Role> roles, String linkIdRegex) {
+        this(serviceId, siteId, name, roles, linkIdRegex, false);
+    }
+
+    public Service(int serviceId, int siteId, String name, Set<Role> roles, String linkIdRegex, boolean disabled) {
```
sunnywu, line 29: is this.disabled = disabled not set?
maybe you need more unit test/assertion to catch this?

src/main/java/com/uid2/shared/model/Service.java
```diff
@@ -73,6 +86,7 @@ public String toString() {
                 ", name='" + name + '\'' +
                 ", roles=" + roles +
                 ", linkIdRegex=" + linkIdRegex +
+                ", disabled=" + disabled +
```
jon8787, line 89: didn't we decide to use lombok at some point?

src/test/java/com/uid2/shared/store/RotatingServiceLinkStoreTest.java
```diff
@@ -62,10 +62,10 @@ public void loadContent_emptyArray_loadsZeroServiceLinks() throws Exception {
     @Test
     public void loadContent_multipleServiceLinksStored_loadsAllServiceLinks() throws Exception {
         JsonArray content = new JsonArray();
-        ServiceLink l1 = addServiceLink(content, "abc123", 1, 123, "Test Service 1", Set.of());
-        ServiceLink l2 = addServiceLink(content, "abc123", 2, 123, "test1", Set.of(Role.MAPPER));
-        ServiceLink l3 = addServiceLink(content, "ghi789", 1, 123, "Test Service 1", Set.of(Role.MAPPER, Role.SHARER));
-        ServiceLink l4 = addServiceLink(content, "jkl1011", 3, 124, "test2", null);
+        ServiceLink l1 = addServiceLink(content, "abc123", 1, 123, "Test Service 1", Set.of(), false);
+        ServiceLink l2 = addServiceLink(content, "abc123", 2, 123, "test1", Set.of(Role.MAPPER), true);
```
jon8787, line 66: we are introducing a mix of enabled/disabled in the test but what tests or assertions are we doing with these?

src/main/java/com/uid2/shared/model/Service.java
Ian-Nara: Thanks. fixed. Is it too granular to be worth testing like assertEquals(true, s.isDisabled()) for each field?

sunnywu: if you missed this earlier then maybe it is worth testing to add a minimal test :)

src/main/java/com/uid2/shared/model/ServiceLink.java
```diff
@@ -83,12 +99,17 @@ public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         ServiceLink serviceLink = (ServiceLink) o;
-        return siteId == serviceLink.siteId && serviceId == serviceLink.serviceId && linkId.equals(serviceLink.linkId)
-                && name.equals(serviceLink.name) && roles.equals(serviceLink.roles);
+        return siteId == serviceLink.siteId 
+            && serviceId == serviceLink.serviceId 
+            && linkId.equals(serviceLink.linkId)
+            && disabled == serviceLink.disabled
```
sunnywu, line 105: this disabled check is repeated in line 108?!

src/test/java/com/uid2/shared/store/RotatingServiceLinkStoreTest.java
jon8787: Where will the tests be that assert that disabling a service link or service, actually disables something?

src/main/java/com/uid2/shared/model/Service.java
Ian-Nara: added

PR #512, opened 2025-08-11
Made refreshFrom a mandatory field and removed field length check when parsing salt files

gmsdelmundo:
`SaltFileParserTest` has been updated such that `refreshFrom` field has a value present in all cases, and `previousSalt` has an empty-string value in the test data

Updated `uid2-shared` locally across the following repos and ran E2Es:
* `uid2-admin`
* `uid2-core`
* `uid2-operator`
* `uid2-optout`
* `uid2-snowflake`

### DIFF
--- a/src/main/java/com/uid2/shared/model/SaltEntry.java
+++ b/src/main/java/com/uid2/shared/model/SaltEntry.java
@@ -12,32 +12,32 @@ public record SaltEntry(
         KeyMaterial currentKey,
         KeyMaterial previousKey
 ) {
-//    @Override
-//    public String toString() {
-//        return "SaltEntry{" +
-//                "id=" + id +
-//                ", hashedId='" + hashedId + '\'' +
-//                ", lastUpdated=" + lastUpdated +
-//                ", currentSalt=<REDACTED>" +
-//                ", refreshFrom=" + refreshFrom +
-//                ", previousSalt=<REDACTED>" +
-//                ", currentKey=" + currentKey +
-//                ", previousKey=" + previousKey +
-//                '}';
-//    }
+    @Override
+    public String toString() {
+        return "SaltEntry{" +
+                "id=" + id +
+                ", hashedId='" + hashedId + '\'' +
+                ", lastUpdated=" + lastUpdated +
+                ", currentSalt=<REDACTED>" +
+                ", refreshFrom=" + refreshFrom +
+                ", previousSalt=<REDACTED>" +
+                ", currentKey=" + currentKey +
+                ", previousKey=" + previousKey +
+                '}';
+    }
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/uid2/shared/store/salt/SaltFileParser.java
```diff
@@ -27,30 +27,23 @@ public SaltEntry[] parseFileLines(String[] saltFileLines, Integer size) {
...
-            // and refreshable UIDs features get rolled out in production. We can remove them one by one as necessary.
-            // AU, 2025/04/28
-            if (fields.length > 3) {
-                refreshFrom = Long.parseLong(fields[3]);
-            }
-            if (fields.length > 4) {
-                previousSalt = fields[4];
-            }
-            if (fields.length > 7) {
+            if (fields.length > 7 && trimToNull(fields[5]) != null && trimToNull(fields[6]) != null) {
```
vishalegbert-ttd, line 42: What does this mean?

src/main/java/com/uid2/shared/model/SaltEntry.java
```diff
@@ -12,32 +12,32 @@ public record SaltEntry(
...
-                ", hashedId='" + hashedId + '\'' +
-                ", lastUpdated=" + lastUpdated +
-                ", currentSalt=<REDACTED>" +
-                ", refreshFrom=" + refreshFrom +
-                ", previousSalt=<REDACTED>" +
-                ", currentKey=" + currentKey +
-                ", previousKey=" + previousKey +
-                '}';
-    }
+//    @Override
```
vishalegbert-ttd, line 15: Delete these commented out functions?

src/main/java/com/uid2/shared/store/salt/SaltFileParser.java
gmsdelmundo: Line 30
```
final String[] fields = line.split(",");
```
is now
```
final String[] fields = line.split(",", -1);
```

---

The second value being `-1` makes it such that if a CSV looks like
```
a,b,c,d,
```
The array will result in
```
["a", "b", "c", "d", ""]
```
instead of
```
["a", "b", "c", "d"]
```

The reason for this is the CSV line is structured where `previousSalt` is the fifth field, and it's an optional value. I wanted to enforce that the field will always exist even if it's an empty string.

But by doing so, if a CSV line looks like the following where the last three fields represents an empty `currentKey`
```
a,b,c,d,,,,
```
It will get parsed as
```
["a", "b", "c", "d", "", "", "", ""]
```
So if we do indeed have more than 7 fields, we still need to check that the mandatory `currentKey` fields (i.e. fields 6 and 7) are not null/empty string

PR #539, opened 2025-09-12
xuy-UID2-6055-upgrade-awsjavasdk-v2-2

clarkxuyang:
Use path style s3 url for e2e test.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
     <groupId>com.uid2</groupId>
     <artifactId>uid2-shared</artifactId>
-    <version>11.1.6-alpha-300-SNAPSHOT</version>
+    <version>11.1.10</version>
     <name>${project.groupId}:${project.artifactId}</name>
     <description>Library for all the shared uid2 operations</description>
     <url>https://github.com/IABTechLab/uid2docs</url>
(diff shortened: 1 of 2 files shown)


src/main/java/com/uid2/shared/cloud/CloudStorageS3.java
```diff
@@ -87,17 +93,23 @@ public CloudStorageS3(String region, String bucket, String s3Endpoint) {
                     .credentialsProvider(credentialsProvider)
                     .region(Region.of(region))
                     .build();
-        } else {
+        } 
+        else {
```
gmsdelmundo, line 97: No need to put else on new line

pom.xml
```diff
@@ -283,10 +283,17 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <version>3.8.0</version>
+                <version>3.12.1</version>
                 <configuration>
                     <source>21</source>
                     <target>21</target>
+                    <annotationProcessorPaths>
```
gmsdelmundo, line 290: Do we need this added?

PR #554, opened 2025-09-24
deprecated methods and make optoout files truly parallel download

lizk886:
https://vertx.io/docs/guides/vertx-5-migration-guide/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
     <groupId>com.uid2</groupId>
     <artifactId>uid2-shared</artifactId>
-    <version>11.1.21-alpha-312-SNAPSHOT</version>
+    <version>11.1.62</version>
     <name>${project.groupId}:${project.artifactId}</name>
     <description>Library for all the shared uid2 operations</description>
     <url>https://github.com/IABTechLab/uid2docs</url>


src/main/java/com/uid2/shared/vertx/CloudSyncVerticle.java
```diff
@@ -232,10 +232,10 @@ private Future<Void> cloudRefresh() {
 
         Promise<Void> refreshPromise = Promise.promise();
         this.isRefreshing = true;
-        vertx.<Void>executeBlocking(blockBromise -> {
+        vertx.executeBlocking(() -> {
             this.cloudRefreshEnsureInSync(refreshPromise, 0);
-            blockBromise.complete();
-        }, ar -> {});
+            return null;
+        });
```
lizk886, line 238: // 4.x
Future<String> fut = vertx.executeBlocking(promise -> promise.complete("result"));

// 5.0
Future<String> fut = vertx.executeBlocking(() -> "result");

src/main/java/com/uid2/shared/vertx/CloudSyncVerticle.java
```diff
@@ -398,15 +394,14 @@ private void cloudDownloadBlocking(Promise<Void> promise, String s3Path) {
...
-            promise.complete();
         } catch (Exception ex) {
             final long cloudDownloadEnd = System.nanoTime();
             final long cloudDownloadTimeMs = (cloudDownloadEnd - cloudDownloadStart) / 1_000_000;
             
             downloadFailureTimer.record(java.time.Duration.ofMillis(cloudDownloadTimeMs));            
             // Be careful as the s3Path may contain the pre-signed S3 token, so do not log the whole path
             LOGGER.error("download error: " + ex.getClass().getSimpleName());
-            promise.fail(new Throwable(ex));
+            throw ex;
```
lizk886, line 404: tell Vert.x that the operation failed, so if ar.succeeded() will have correct status if needed

clarkxuyang: is the original wrapper a new Throwable(ex) on top of ex is because of 
// Be careful as the s3Path may contain the pre-signed S3 token, so do not log the whole path ?

clarkxuyang (approved):

src/main/java/com/uid2/shared/vertx/CloudSyncVerticle.java
lizk886: revert to old method for safety

PR #610, opened 2026-04-01
UID2-1750: fix Lombok scope and add MIT license attribution

sunnywu:
## Summary
- Change Lombok from default `compile` scope to `provided` + `optional` so it no longer propagates as a transitive dependency to downstream services
- Register Lombok as an explicit `annotationProcessorPath` in `maven-compiler-plugin` for Java 9+ module system compatibility
- Add `NOTICE` file with Lombok copyright acknowledgment (compile-time only note)
- Add `THIRD-PARTY-LICENSES/lombok-LICENSE.txt` with full MIT license text

## Why
Lombok defaulting to `compile` scope caused it to leak transitively into all consumers of uid2-shared. This is architecturally wrong — Lombok is a build tool, not a runtime library — and means downstream JARs/Docker images bundle `lombok.jar`, triggering MIT attribution obligations in those projects. See [UID2-1750](https://thetradedesk.atlassian.net/browse/UID2-1750).

## Test plan
- [x] `mvn compile` passes (198 source files, 0 errors)
- [x] `mvn test` passes (439 tests, 0 failures)
- [x] uid2-admin compiles and all 698 tests pass with companion PR IABTechLab/uid2-admin#615

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -215,7 +215,6 @@
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
             <version>1.18.34</version>
-            <scope>provided</scope>
             <optional>true</optional>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -215,6 +215,8 @@
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
             <version>1.18.34</version>
+            <scope>provided</scope>
+            <optional>true</optional>
         </dependency>
         <dependency>
             <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -215,17 +222,14 @@
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
             <version>1.18.34</version>
+            <scope>provided</scope>
```
jon8787: curious, do we actually need `provided` given there is no runtime and `optional` is there?

sunnywu: what claude said:

uid2-shared is a library (not a leaf project), so the relevant scoping concerns are different:                                                       
                                                                                                                                                       
  - optional — correct and important here. It prevents downstream projects that depend on uid2-shared from inheriting lombok as a transitive           
  dependency.                                                                                                                                          
  - provided — Jon's question is fair. For a library JAR, Maven doesn't bundle dependencies regardless of scope (that's only for fat JARs). And since  
  optional already stops transitive propagation, provided is technically redundant from a Maven behavior standpoint.                                   
  
  So for uid2-shared specifically: optional alone is sufficient to achieve the desired effect. provided doesn't add any functional difference — it's   
  redundant, though harmless.

jon8787 (approved):

jon8787 (approved):

PR #613, opened 2026-06-03
UID2-4739: use DefaultCredentialsProvider for S3 clients in CloudStorageS3

sophia-chen-ttd:
## Summary

- `CloudStorageS3` explicitly constructed `WebIdentityTokenFileCredentialsProvider`, reading `AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE` env vars directly — tying it hard to OIDC/IRSA
- Replaced with `DefaultCredentialsProvider.create()`, which automatically resolves credentials via the standard AWS SDK credential chain (IRSA, EKS Pod Identity, instance profiles, env vars, etc.)
- The `AWS_WEB_IDENTITY_TOKEN_FILE` env var read is removed as it was only used to construct the old provider

## Why

`WebIdentityTokenFileCredentialsProvider` only works with IRSA (OIDC-based service account annotation). With EKS Pod Identity (the newer, preferred mechanism), credentials are injected differently and this explicit provider will fail to pick them up. `DefaultCredentialsProvider` handles both IRSA and EKS Pod Identity transparently, and will continue to work with any future AWS credential mechanism without code changes.

## ⚠️ Historical context — read before merging

The original code contained this comment (now removed):

> In theory `new InstanceProfileCredentialsProvider()` or even omitting credentials provider should work, but for some unknown reason it doesn't. The credential it provides look realistic, but are not valid. After a lot of experimentation and help of Abu Abraham and Isaac Wilson the only working solution we've found was to explicitly extract env vars populated by the service account from the role and to manually set it on the credentials provider.

This failure was most likely caused by an **old AWS SDK v2 version** that did not fully auto-detect IRSA credentials via the default chain. Early SDK v2 releases had gaps in `WebIdentityTokenCredentialsProvider` auto-detection, requiring explicit construction from env vars. This was a known SDK issue fixed in later versions.

Modern AWS SDK v2 includes `WebIdentityTokenCredentialsProvider` in the default chain and detects `AWS_WEB_IDENTITY_TOKEN_FILE` automatically. `DefaultCredentialsProvider` also handles EKS Pod Identity via the container credentials endpoint (`AWS_CONTAINER_CREDENTIALS_FULL_URI`).

**This change must be explicitly tested** in a non-prod EKS environment before merging — the historical comment is a warning that this code path failed silently once. Do not assume the theoretical fix works without verification.

## Test plan

- [ ] Unit tests pass (`mvn test`)
- [ ] **Deploy to a non-prod environment and confirm S3 operations succeed** — this is the critical test given the historical failure above
- [ ] Confirm existing IRSA-based environments continue to work (`DefaultCredentialsProvider` still picks up IRSA env vars via its `WebIdentityTokenCredentialsProvider` step in the chain)
- [ ] After EKS Pod Identity migration: confirm S3 operations succeed with Pod Identity credentials

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/main/java/com/uid2/shared/cloud/CloudStorageS3.java
```diff
@@ -74,15 +74,9 @@ public CloudStorageS3(String accessKeyId, String secretAccessKey, String region,
...
-        // In theory `new InstanceProfileCredentialsProvider()` or even omitting credentials provider should work,
-        // but for some unknown reason it doesn't. The credential it provides look realistic, but are not valid.
-        // After a lot of experimentation and help of Abu Abraham and Isaac Wilson the only working solution we've
-        // found was to explicitly extract env vars populated by the service account from the role and to
-        // manually set it on the credentials provider.
-        WebIdentityTokenFileCredentialsProvider credentialsProvider = WebIdentityTokenFileCredentialsProvider.builder()
-                .roleArn(System.getenv("AWS_ROLE_ARN"))
-                .webIdentityTokenFile(Paths.get(System.getenv("AWS_WEB_IDENTITY_TOKEN_FILE")))
-                .build();
+        // DefaultCredentialsProvider supports IRSA (WebIdentityTokenFile), EKS Pod Identity,
```
sophia-chen-ttd, line 77: **Claude Review:** The old comment is removed entirely, losing useful institutional memory. Consider retaining a condensed note documenting *why* the explicit provider was used historically and *why* the default chain is safe now — e.g.:

```java
// DefaultCredentialsProvider resolves via the standard AWS chain: env vars → system properties
// → web identity token (IRSA) → EKS Pod Identity → instance profile. Prefer this over hard-coding
// WebIdentityTokenFileCredentialsProvider, which only covered IRSA and broke with old SDK versions.
```

src/main/java/com/uid2/shared/cloud/CloudStorageS3.java
```diff
@@ -74,15 +74,9 @@ public CloudStorageS3(String accessKeyId, String secretAccessKey, String region,
...
+        // instance profile, and all other standard AWS credential mechanisms automatically.
+        DefaultCredentialsProvider credentialsProvider = DefaultCredentialsProvider.create();
```
sophia-chen-ttd, line 79: **Claude Review:** `DefaultCredentialsProvider` is lazy — it does not resolve credentials until the first API call. If the chain falls through to the EKS Pod Identity or IMDS provider, the first real S3 call will block on an HTTP metadata fetch, surfacing misconfiguration at request time rather than startup. Consider calling `credentialsProvider.resolveCredentials()` in the constructor (in a try-catch that logs a warning) to fail fast with a clear message.

sophia-chen-ttd, line 79: **Claude Review:** No unit or integration tests exist for `CloudStorageS3`. Given the PR body explicitly requires non-prod environment testing before merge, consider adding at minimum a constructor-level unit test verifying the 3-arg constructor builds without throwing when `AWS_ROLE_ARN` / `AWS_WEB_IDENTITY_TOKEN_FILE` are absent (the old code would NPE at construction; the new code must not).

sophia-chen-ttd (commented):
The swap to `DefaultCredentialsProvider` is the correct direction: it supports IRSA (WebIdentityTokenFile), EKS Pod Identity (container credentials endpoint), instance profile, and all other standard chain entries transparently. The old comment's 'unknown reason' failure was almost certainly an early AWS SDK v2 version that predated reliable IRSA support in the default chain — the PR's explanation is credible. The change is minimal and low-risk, but removal of the historical warning, lack of credential test coverage, and resource-lifecycle subtleties are worth calling out.

src/main/java/com/uid2/shared/cloud/CloudStorageS3.java
sophia-chen-ttd: Dismissing: the PR description's [Historical context section](https://github.com/IABTechLab/uid2-shared/pull/613) already captures the full story (old SDK version gap, why it failed, why the default chain is safe now). That context belongs in git history, not in an inline comment that will grow stale as the SDK evolves. The current one-liner is sufficient.

src/main/java/com/uid2/shared/cloud/CloudStorageS3.java
sophia-chen-ttd: Dismissing: calling `resolveCredentials()` eagerly in the constructor is not a good fit here. AWS SDK v2 credential providers issue temporary STS tokens that auto-refresh — an eager check at startup doesn't prevent expiry issues later, gives a false sense of security, and would break any unit test that constructs `CloudStorageS3` without real credentials. The old `WebIdentityTokenFileCredentialsProvider` was also lazy at construction. The non-prod deployment test in the test plan is the right validation path.

sophia-chen-ttd: Added test

mcollins-ttd (approved):

